PR #5, opened 2023-05-30
Make interface/models.py run under Python 3.7

kmaziarz:
The generic single-step model interface is not used yet by the search code, and thus also wasn't covered by CI. It turns out it didn't fully work, due to (1) `pydantic` not being present in the environment, and (2) `dict`/`list` not fully working under Python 3.7 in this context. This PR addresses both points and adds a minimal test to cover `models.py`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,3 +17,6 @@ dependencies:
   - pytest
   - pytest-cov
   - pre-commit
+  - pip:
+    # Optional dependency for the single-step model interface
+    - pydantic>=1.10.5
--- a/environment.yml
+++ b/environment.yml
@@ -19,4 +19,4 @@ dependencies:
   - pre-commit
   - pip:
     # Optional dependency for the single-step model interface
-    - pydantic>=1.10.5
+    - pydantic>=1.10.5  # earlier versions had a bug involving `default_factory` (see https://github.com/pydantic/pydantic/issues/5065)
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -17,3 +17,6 @@ dependencies:
   - pytest
   - pytest-cov
   - pre-commit
+  - pip:
+    # Optional dependency for the single-step model interface
+    - pydantic>=1.10.5
```
AustinT, line 22: I know this version requirement is copied in the internal repo, but is there any particular reason why it is required? Would anything bad happen if we unpinned it?

syntheseus/tests/interface/test_models.py
```diff
@@ -0,0 +1,11 @@
+import math
+
+import numpy as np
+
+from syntheseus.interface.models import Prediction
+
+
+def test_prediction():
```
AustinT, line 8: Is this test just there to make sure imports from `models.py` work? If so, maybe add this as a comment or TODO item in the code to remind our future selves 😛

AustinT (approved):
I left a few comments but otherwise looks good to me!

environment.yml
kmaziarz: Yeah, `1.10.2` has a bug related to `default_factory` (see [here](https://github.com/pydantic/pydantic/issues/5065)), so we wouldn't go to `<1.10` (we could maybe move the pin a few patch versions back, not sure exactly where between `1.10.2` and `1.10.5` the bug was fixed, but it doesn't seem worthwhile).

PR #8, opened 2023-06-26
Add initial code for PDVN training

AustinT:
This PR adds a file `search/algorithms/pdvn.py` which contains code to generate training data for PDVN using a version of AND/OR MCTS + postprocessing code to extract training data for policies and value functions.


syntheseus/search/algorithms/pdvn.py
```diff
@@ -0,0 +1,372 @@
+"""
+Contains code related to the "Retrosynthetic Planning with Dual Value Networks" algorithm.
+
+See: https://arxiv.org/abs/2301.13755
+
+In particular, this file contains code for the And/OR-MCTS algorithm used to make training data for PDVN,
```
kmaziarz, line 6: nit: for consistency
```suggestion
In particular, this file contains code for the AND/OR-MCTS algorithm used to make training data for PDVN,
```

syntheseus/search/algorithms/pdvn.py
```diff
@@ -0,0 +1,372 @@
...
+        the same logic is used to select a child at an OrNode (selecting a single reaction). However,
+        at an AndNode the logic is different: a single OrNode child is chosen using the logic in
+        section 3.2 of Liu et al 2023, which states:
+
+            To select a child molecule node, we prioritize molecules that have not been expanded;
+            if none are available, we choose ones that have not been solved. If molecule nodes are
+            either all expanded or all solved, we randomly select one of them.
+
+        The logic below follows the paragraph above, disregarding that the final clause
+        will only be reached if the childre are all expanded *and* all solved, rather than *or*
```
kmaziarz, line 136:
```suggestion
        will only be reached if the children are all expanded *and* all solved, rather than *or*
```

syntheseus/search/algorithms/pdvn.py
```diff
@@ -0,0 +1,372 @@
...
+                children_to_choose_from = children
+            output = [self.random_state.choice(children_to_choose_from)]
+        else:
+            raise TypeError(f"Unexpected node type: {type(node)}")
+
+        # Ensure only 1 node was chosen
+        assert len(output) == 1, "PDVN_MCTS can only choose 1 successor to visit."
+        return output  # type: ignore[return-value]  # mypy thinks it could be any graph node due to super() call
+
+    def _get_leaf_node_reward(self, node: OrNode, graph: AndOrGraph) -> tuple[float, float]:  # type: ignore[override]
```
kmaziarz, line 165: This is fine for now, but eventually should we make the reward type generic @AustinT?

syntheseus/search/algorithms/pdvn.py
```diff
@@ -0,0 +1,372 @@
...
+    def _set_and_node_costs(self, and_nodes: Sequence[AndNode], graph: AndOrGraph) -> None:
+        """Helper function to compute the cost of each AndNode using the `and_node_cost_fn`."""
+        costs = self.and_node_cost_fn(and_nodes, graph=graph)
+        assert len(costs) == len(and_nodes)
+        for node, cost in zip(and_nodes, costs):
+            node.data["pdvn_reaction_cost"] = cost
+
+    def choose_successors_to_visit(
+        self,
+        node: ANDOR_NODE,  # type: ignore[override]  # AND/OR vs any graph node
```
kmaziarz, line 120: Fine for now, but @AustinT do you think we should later improve the generics to resolve this? I think this could be done but it might introduce extra complexity (e.g. yet another type variable in the base class)...

syntheseus/search/algorithms/pdvn.py
```diff
@@ -0,0 +1,372 @@
...
+            node.data["pdvn_mcts_v_syn"] * node.data["pdvn_mcts_v_cost"]
+            + (1 - node.data["pdvn_mcts_v_syn"]) * self.c_dead
+        )
+        node.num_visit += 1
+
+
+def pdvn_min_cost_update(node: ANDOR_NODE, graph: AndOrGraph):
+    """
+    Update function to compute "pdvn_min_syn_cost" for a given node.
+    the pdvn_min_syn_cost is the cost of the minimum synthesis route to synthesize
```
kmaziarz, line 271:
```suggestion
    The pdvn_min_syn_cost is the cost of the minimum synthesis route to synthesize
```

syntheseus/search/node_evaluation/common.py
```diff
@@ -37,15 +38,17 @@ def _evaluate_nodes(self, nodes, graph=None) -> Sequence[float]:
         return [-v for v in super()._evaluate_nodes(nodes, graph)]
 
 
-class ReactionModelProbPolicy(ReactionModelBasedEvaluator[MolSetNode]):
-    """Evaluator that uses the reactions' probability to form a policy (useful for OR-MCTS)."""
+class ReactionModelProbPolicy(ReactionModelBasedEvaluator[Union[MolSetNode, AndNode]]):
+    """Evaluator that uses the reactions' probability to form a policy (useful for OR-MCTS and And/OR-MCTS)."""
```
kmaziarz, line 42: nit: maybe just
```suggestion
    """Evaluator that uses the reactions' probability to form a policy (useful for MCTS)."""
```

syntheseus/search/node_evaluation/common.py
```diff
@@ -37,15 +38,17 @@ def _evaluate_nodes(self, nodes, graph=None) -> Sequence[float]:
...
-        assert len(parents) == 1, "Graph must be a tree"
-
-        return graph._graph.edges[parents[0], node]["reaction"]
+    def _get_reaction(self, node: Union[MolSetNode, AndNode], graph) -> BackwardReaction:
+        if isinstance(node, MolSetNode):
+            parents = list(graph.predecessors(node))
+            assert len(parents) == 1, "Graph must be a tree"
+            return graph._graph.edges[parents[0], node]["reaction"]
+        else:
+            return node.reaction
```
kmaziarz, line 54: Maybe it doesn't hurt to be a bit safer just in case
```suggestion
        if isinstance(node, MolSetNode):
            parents = list(graph.predecessors(node))
            assert len(parents) == 1, "Graph must be a tree"
            return graph._graph.edges[parents[0], node]["reaction"]
        elif isinstance(node, AndNode):
            return node.reaction
        else:
            raise ValueError(f"ReactionModelProbPolicy does not support nodes of type {type(node)}")
```

syntheseus/tests/search/algorithms/test_pdvn.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+@pytest.fixture
+def policy(
+    rxn_cost_fn,  # noqa: F811  # doesn't understand pytest's use of rxn_cost_fn as fixture
+) -> DictRxnCost:
+    return DictRxnCost(
+        rxn_smiles_to_cost={rxn: math.exp(-c) for rxn, c in rxn_cost_fn.rxn_to_cost.items()},
+        default=0.4,
+    )
```
kmaziarz, line 28: Huh, weird - is it `flake8` that is complaining about `rxn_cost_fn` being seemingly unused?

syntheseus/tests/search/algorithms/test_pdvn.py
```diff
@@ -0,0 +1,428 @@
...
+            value_function_cost=mol_cost_estimate,
+        )
+        assert output_graph.reaction_smiles_counter() == Counter(rxns_step2)  # type: ignore[attr-defined]
+
+        # Various checks on root node
+        assert output_graph.root_node.num_visit == 4
+        assert math.isclose(
+            output_graph.root_node.data["pdvn_mcts_v_syn"],
+            2.6 / 4,
+        ), "v_syn should be 0.4 average(0, 0.8, 0.8, 1.0)"
```
kmaziarz, line 179: Value in the error message doesn't match what is being checked
```suggestion
        assert math.isclose(
            output_graph.root_node.data["pdvn_mcts_v_syn"],
            2.6 / 4,
        ), "v_syn should be 0.65 average(0, 0.8, 0.8, 1.0)"
```

kmaziarz (approved):
LGTM (remember to add a CHANGELOG entry before merging).

fiberleif:
> remember to add a CHANGELOG entry before merging

Added

syntheseus/tests/search/algorithms/test_pdvn.py
```diff
@@ -0,0 +1,428 @@
...
+        assert math.isclose(
+            output_graph.root_node.data["pdvn_mcts_v_cost"],
+            10 / 4,
+        ), "v_cost should be 1.5 average(0, 3, 3, 3, 1)"
```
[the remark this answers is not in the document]
fiberleif: After running pytest locally, l found the right values are: `average(0, 3, 3, 4)`

syntheseus/search/algorithms/pdvn.py
AustinT: Yes I agree. I considered doing it with this PR but didn't want to introduce a lot of additional complexity for what might end up being a "one off" algorithm.

syntheseus/search/algorithms/pdvn.py
AustinT: Yes, the generics for graph types are a big weakness of this package. I really wanted to resolve this before the initial release but couldn't figure out how to do so. If you want to look into it you are very welcome to do so!

syntheseus/tests/search/algorithms/test_pdvn.py
AustinT: Yes, that's my understanding. Syntactically it is unused, it just ends up being used by `pytest` but `flake8` doesn't "know" about this.

syntheseus/search/node_evaluation/common.py
```diff
@@ -37,15 +38,17 @@ def _evaluate_nodes(self, nodes, graph=None) -> Sequence[float]:
         return [-v for v in super()._evaluate_nodes(nodes, graph)]
 
 
-class ReactionModelProbPolicy(ReactionModelBasedEvaluator[MolSetNode]):
-    """Evaluator that uses the reactions' probability to form a policy (useful for OR-MCTS)."""
+class ReactionModelProbPolicy(ReactionModelBasedEvaluator[Union[MolSetNode, AndNode]]):
```
AustinT, line 41: Do you think it is better to put this into a single class with a node type check instead of having two separate classes?

AustinT (commented):
New changes look good Guoqing. I checked that the tests pass when I run them locally on my computer.

syntheseus/search/chem.py
```diff
@@ -15,8 +15,10 @@ class ReactionMetaData(TypedDict, total=False):
     cost: float
     template: str
     source: str  # any explanation of the source of this reaction
+    probability: float  # probability used for multi-step search
```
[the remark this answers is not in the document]
AustinT: I just realized it was also used in a previous PR. Still, @kmaziarz or @fiberleif do you have any opinion about whether "score" or "probability" is a better name?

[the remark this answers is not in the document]
fiberleif: Yes, l added a `probability` attribute here since l found this attribute will be used in some other places we're using (e.g., ReactionModelBasedEvaluator, etc)

syntheseus/search/node_evaluation/common.py
fiberleif: Are you suggesting that I replace `Union[MolSetNode, AndNode]` with a single `BaseGraphNode` for typing? Or did you mean something else? I am using a single ReactionModelProbPolicy class to handle both MolSetNode in OR-MCTS and AndNode in PDVN-MCTS here

syntheseus/search/chem.py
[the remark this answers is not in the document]
AustinT: Fair, but I think `probability` is an loaded term with a lot of connotations (e.g. normalization) which the values of `probability` may not satisfy. This is why I originally used the word `score`.

syntheseus/search/node_evaluation/common.py
AustinT: In the current branch these are [two separate classes](https://github.com/microsoft/syntheseus/blob/8516d3108070f97a3e31f185fb6c2e3c38ff8d74/syntheseus/search/node_evaluation/common.py#L27-L51), so my question is why merge them into one class?

kmaziarz: The other class on `main` is negative log cost, whereas here we use probabilities. Would you suggest to now have _three_ classes for each of the usecases?

AustinT: Ah, you are right (I didn't read this carefully). Let's just merge them then as Guoqing suggests.

syntheseus/search/chem.py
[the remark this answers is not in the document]
fiberleif: Understood! Perhaps for this PR, we can continue to use the term "probability" to maintain compatibility with previous PRs. However, if we ultimately determine that "score" is more appropriate, we can create a new PR to fully implement the change.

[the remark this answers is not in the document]
kmaziarz: Yeah, as you both notice, `probability` is already used to store single-step model probability in `SyntheseusBackwardReactionModel`. We could switch to `score` in a follow-up PR if we feel this is better, although I did check in the past that for all single-step models what is placed under `probability` is actually in `[0, 1]`, so it seems reasonably consistent with the definition of "probability"...

PR #9, opened 2023-06-29
Add code to extract routes in order found.

AustinT:
1) Add code to route extraction (simple wrapper over _iter_top_routes)

2) Added tests in test_route_extraction.py

3) While writing the tests I sort of forgot what my test graph look like, so I made an ASCII drawing in conftest.py.

### DIFF
--- a/syntheseus/search/analysis/route_extraction.py
+++ b/syntheseus/search/analysis/route_extraction.py
@@ -189,3 +189,40 @@ def min_cost_routes(
             break
         else:
             yield route
+
+
+def _route_time_cost(nodes, graph) -> float:
+    """
+    Cost function for routes that is the maximum timestamp of any node in the route
+    (or inf if it is not solved).
+    """
+    if _route_has_solution(nodes, graph):
+        return max(n.creation_time.timestamp() for n in nodes)
+    else:
+        return math.inf
+
+
+def _route_time_partial_cost(nodes, _) -> float:
+    """Partial cost version of above. It is a lower bound to the true cost."""
+    if all(n.has_solution for n in nodes):
+        return max(n.creation_time.timestamp() for n in nodes)
+    else:
+        return math.inf
+
+
+def iter_routes_in_order_found(
+    graph: RetrosynthesisSearchGraph, max_routes: int
+) -> Iterator[Collection[BaseGraphNode]]:
+    """
+    Yield all routes from "graph" in the order they were found.
+    Graph can be AND/OR or MolSet graph. Order is determined by "creation_time".
+    """
+
+    for _, r in _iter_top_routes(
+        graph=graph,
+        cost_fn=_route_time_cost,
+        cost_lower_bound=_route_time_partial_cost,
+        max_routes=max_routes,
+        yield_partial_routes=False,
+    ):
+        yield r
--- a/syntheseus/tests/search/analysis/test_route_extraction.py
+++ b/syntheseus/tests/search/analysis/test_route_extraction.py
@@ -10,14 +10,14 @@ from __future__ import annotations
 import numpy as np
 import pytest
 
-from syntheseus.search.analysis.route_extraction import min_cost_routes
+from syntheseus.search.analysis.route_extraction import iter_routes_in_order_found, min_cost_routes
 from syntheseus.search.graph.and_or import AndNode, AndOrGraph
 from syntheseus.search.graph.molset import MolSetGraph
 from syntheseus.search.graph.route import SynthesisGraph
 from syntheseus.tests.search.analysis.conftest import set_uniform_costs
 
 
-def test_correct_routes_andor(
+def test_correct_min_cost_routes_andor(
     andor_graph_non_minimal: AndOrGraph,
     andor_tree_non_minimal: AndOrGraph,
     minimal_synthesis_graph: SynthesisGraph,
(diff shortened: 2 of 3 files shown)


syntheseus/search/analysis/route_extraction.py
```diff
@@ -189,3 +189,40 @@ def min_cost_routes(
...
+    else:
+        return math.inf
+
+
+def iter_routes_in_order_found(
+    graph: RetrosynthesisSearchGraph, max_routes: int
+) -> Iterator[Collection[BaseGraphNode]]:
+    """
+    Yield all routes from "graph" in the order they were found.
+    Graph can be AND/OR or MolSet graph. Order is determined by "creation_time".
```
kmaziarz, line 218: nit: If you are going to put something around variable names or other symbols, then I think the standard thing is to use backticks like [here](https://github.com/microsoft/syntheseus/blob/e255b0559286aee34584742dc980e68b60f6cae2/syntheseus/search/graph/message_passing/run.py#L34), e.g.
```suggestion
    Yield all routes from `graph` in the order they were found.
    Graph can be AND/OR or MolSet graph. Order is determined by `creation_time`.
```
But while some parts of `syntheseus` use backticks, other parts don't use anything, so it's not too consistent at the moment.

syntheseus/tests/search/analysis/test_route_extraction.py
```diff
@@ -70,6 +70,56 @@ def test_correct_routes_molset(
...
+def test_correct_routes_andor_time_order(
+    andor_graph_non_minimal: AndOrGraph,
+    andor_tree_non_minimal: AndOrGraph,
+    minimal_synthesis_graph: SynthesisGraph,
+) -> None:
+    """
+    Test that the correct routes are extracted from an AND/OR graph
+    in the order that they should be found.
+
+    To to this, we check that the number of routes found is correct
```
kmaziarz, line 82: typo:
```suggestion
    To do this, we check that the number of routes found is correct
```

syntheseus/search/analysis/route_extraction.py
```diff
@@ -189,3 +189,40 @@ def min_cost_routes(
...
+
+def _route_time_partial_cost(nodes, _) -> float:
+    """Partial cost version of above. It is a lower bound to the true cost."""
+    if all(n.has_solution for n in nodes):
+        return max(n.creation_time.timestamp() for n in nodes)
+    else:
+        return math.inf
+
+
+def iter_routes_in_order_found(
```
kmaziarz, line 213: nit: This is very minor but I noticed that `min_cost_routes` and `iter_routes_in_order_found` are similar and both return an iterator over routes, while in a sense they follow different "naming schemes"?

kmaziarz (approved):
LGTM, I just found one typo and left a few general thoughts about code style.

syntheseus/search/analysis/route_extraction.py
AustinT: Yeah, I haven't been very consistent, my bad. Which would you prefer? Backticks or quotes?

syntheseus/search/analysis/route_extraction.py
AustinT: Good observation. In hindsight I think `min_cost_routes` should be named `iter_min_cost_routes`, but I didn't want to make breaking changes.

I considered naming the new method `first_routes` instead of `iter_routes_in_order_found`, but I thought that was a bit confusing because it sounds a bit like it returns just the first route.

If you can think of a better name I am happy to rename it, otherwise we can later rename `min_cost_routes` to `iter_min_cost_routes`.

syntheseus/search/analysis/route_extraction.py
kmaziarz: I personally use backticks quite consistently (and so this is what you'll find in e.g. single-step code), but usually don't enforce it on others too aggressively 🙂 After a quick look, there are cases in `syntheseus` that use backticks, quotes, or nothing at all, but maybe if we decide on what we prefer then this will slowly converge one way or another in future PRs.

I think this stuff matters more when one uses an automated tool to convert docstrings into documentation, in which case backticks are likely to be rendered specially. Otherwise it's maybe slightly nicer to read as mentally one knows when to expect a variable/symbol.

syntheseus/search/analysis/route_extraction.py
kmaziarz: > In hindsight I think min_cost_routes should be named iter_min_cost_routes, but I didn't want to make breaking changes.

Yup, that's what I was thinking (or `iter_routes_[something]` but maybe that wouldn't sound natural). Regarding breaking changes, they should be fine given a small number of "users" depending on `syntheseus` downstream (and internal code has `syntheseus` pinned to a particular commit so it won't break by itself), and it's also better sooner than later 😄 But we might find other things worth renaming, so we may want to bundle these together in a PR that purely deals with renaming, and keep this one non-breaking.

syntheseus/search/analysis/route_extraction.py
AustinT: Sounds good. I'm happy with this approach.

syntheseus/search/analysis/route_extraction.py
AustinT: Let's just rename this now and accept breaking changes. I pushed a change which does this.

PR #10, opened 2023-07-07
Pin pydantic version to 1.*

kmaziarz:
Recently, pydantic 2.0 was released, bringing backward incompatible changes, and breaking our CI:
```python
from pydantic.generics import GenericModel
E   ImportError: cannot import name 'GenericModel' from 'pydantic.generics'
```

In the future we can look into bumping the `pydantic` requirement, but for now we can stick to just pinning `pydantic` to `1.*` as done in this PR.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,4 +19,4 @@ dependencies:
   - pre-commit
   - pip:
     # Optional dependency for the single-step model interface
-    - pydantic>=1.10.5  # earlier versions had a bug involving `default_factory` (see https://github.com/pydantic/pydantic/issues/5065)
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - numpy  # required for some array computing
   - rdkit  # required for molecule objects
   - networkx
-  # Optional dependencies for visuzalization
+  # Optional dependencies for visualization
```
AustinT, line 13: Thanks for catching this typo!

AustinT (approved):
Looks good to me!

fiberleif (approved):
LGTM!

PR #11, opened 2023-07-10
Pin zipp<3.16

kmaziarz:
Recently, a library called `zipp` (which we don't depend on directly, but it seems something related to our pre-commit does) released version `3.16`, which dropped support for Python 3.7. It seems that `zipp`'s Python requirement is not correctly picked during installation (see jaraco/zipp#100); this makes our Python 3.7 CI fail as it installs the new version.

In this PR, I pin `zipp<3.16` to get the CI on `main` to pass again. This is not ideal as we don't even depend on `zipp` directly, but I hope it's temporary, and we can remove it once the community sorts things out.


AustinT (approved):
Looks good to me. We can take it out later.

fiberleif (approved):
LGTM!

PR #12, opened 2023-07-11
Avoid running the security pipeline on every PR

kmaziarz:
The security pipeline defined in `azure_pipelines/code-security-analysis.yml` is scheduled to run weekly as required by Microsoft policy. However, it also seems to run on all PRs, which is unnecessary. This PR changes the pipeline's configuration to only do the scheduled runs and not run as part of CI.

### DIFF
--- a/.github/azure_pipelines/code-security-analysis.yml
+++ b/.github/azure_pipelines/code-security-analysis.yml
@@ -5,6 +5,9 @@ schedules:
     include:
     - main
 
+pr: none
+trigger: none
+
 pool:
  vmImage: 'windows-latest'
 


AustinT (approved):
I don't see any problems in this PR, but I also have no way to check that this accomplishes what we intend it to accomplish. If you have looked into it then I will trust you Kris!

kmaziarz:
> I don't see any problems in this PR, but I also have no way to check that this accomplishes what we intend it to accomplish. If you have looked into it then I will trust you Kris!

I can't be 100% sure, but if you look at the checks ran on this PR, the security pipeline (called `microsoft.syntheseus`) is no longer there, while it was there on all previous PRs, so hopefully my changes are doing the right thing 🙂

PR #13, opened 2023-07-11
Add option to stop search once a solution is found

AustinT:
*Rationale*: Even though our main use case for `syntheseus` so far has been to run search for a fixed amount of time and analyze all routes in the resulting search graph, there are some scenarios where we only care about whether or not a route is found (e.g. benchmarking). This is pretty simple to check so I thought it is worth supporting.

*Implementation*: previously all algorithms called `time_limit_reached` each iteration to decide whether to stop early. I just augmented this function to also check for stopping once a solution is found. This involved:

1. Adding a `stop_on_first_solution` kwarg to the base algorithm class (defaults to False, which is the most sensible default in my opinion)
2. Adding logic to `time_limit_reached` which checks `self.stop_on_first_solution and graph.root_node.has_solution`. This required adding a `graph` argument to the function (previously it took no arguments), causing changes in all algorithm files.
3. Because of this added functionality, `time_limit_reached` no longer seemed like a descriptive name for the function, so I changed it to `should_stop_search`
4. I added a test to `syntheseus/tests/search/algorithms/test_base.py` which tests the effectiveness of this kwarg for all algorithms (because algorithm-specific tests all subclass the base test)

*Rejected alternative*: although it would be appealing to generalize this to "stop after n solutions" or "stop after n diverse solutions" this is not simple or quick to check so I don't think we should support it out-of-the-box. If users want this then they could implement it themselves in a subclass.

### DIFF
--- a/syntheseus/tests/search/algorithms/test_base.py
+++ b/syntheseus/tests/search/algorithms/test_base.py
@@ -447,6 +447,7 @@ class BaseAlgorithmTest(abc.ABC):
         time_limit_s: float,
         limit_iterations: int = 10_000,
         max_routes: int = 100,
+        **kwargs,
     ) -> list[SynthesisGraph]:
         """Utility function to run an algorithm and extract routes."""
 
@@ -456,6 +457,7 @@ class BaseAlgorithmTest(abc.ABC):
             mol_inventory=task.inventory,
             limit_iterations=limit_iterations,
             time_limit_s=time_limit_s,
+            **kwargs,
         )
         output_graph, _ = alg.run_from_mol(task.target_mol)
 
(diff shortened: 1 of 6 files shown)


AustinT:
> @AustinT please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree

kmaziarz:
> there are some scenarios where we only care about whether or not a route is found (e.g. benchmarking)

Can you elaborate on this usecase?

syntheseus/tests/search/algorithms/test_base.py
```diff
@@ -499,3 +501,23 @@ def test_found_routes2(self, retrosynthesis_task2: RetrosynthesisTask) -> None:
         for incorrect_route in retrosynthesis_task2.incorrect_routes.values():
             route_matches = [incorrect_route == r for r in route_objs]
             assert not any(route_matches)
+
+    def test_stop_on_first_solution(self, retrosynthesis_task1: RetrosynthesisTask) -> None:
+        """
+        Test that `stop_on_first_solution` really does stop the algorithm once a solution is found.
+
+        The test for this is to run the same search as in `test_found_routes1` but with
+        stop_on_first_solution=True. This should find exactly one route for this problem.
```
kmaziarz, line 510: nit
```suggestion
        `stop_on_first_solution=True`. This should find exactly one route for this problem.
```

kmaziarz (approved):
LGTM!

PR #15, opened 2023-07-26
Release single-step model environments

kmaziarz:
This PR releases instructions for building an appropriate environment for each of the supported single-step models. I've managed to unify all models apart from GLN into a shared base (containing mostly just `python`, `torch` and `rdkit`) which each model type extends with a handful of model-specific dependencies. One last piece still missing are checkpoint files for each model type.

For now, the setup instructions live under `reaction_prediction/environments`. Longer term we could consider forking the external single-step model repositories and adding simple `pip`-installability to them, which could allow us to streamline the setup steps further through installation extras, e.g. be able to do something like `pip install syntheseus[retro-knn]`. Currently the dependencies in the environments are mostly pinned for reproducibility, but if we go with the installation extras approach then we can unpin them wherever possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,20 @@
 [![License](https://img.shields.io/badge/license-MIT-green.svg)](LICENSE)
 
 Syntheseus is a package for retrosynthetic planning.
-It contains implementations of common search algorithms
-and a simple API to wrap custom reaction models and write
-custom algorithms.
+It contains implementations of common search algorithms, a simple API to wrap custom reaction models and write
+custom algorithms, and wrappers for many state-of-the-art reaction models from the literature.
 It is meant to allow for simple benchmarking of the components
 of retrosynthesis algorithms.
 
-## Installation
+## Setup
 
-Currently `syntheseus` is not hosted on PyPI
-(although this will likely change in the future).
-To install, please run:
+We support two installation modes:
+- *core installation* not tied to a specific reaction model allows you to build and benchmark your own models or search algorithms
+- *full installation* backed by one of the supported models allows you to perform end-to-end retrosynthetic search
+
+For full installation we currently support the following reaction models: Chemformer, LocalRetro, MEGAN, MHNreact, RetroKNN and RootAligned SMILES; see [here](syntheseus/reaction_prediction/environments/README.md) for detailed setup instructions.
+
+For core installation simply run
 
 ```bash
 # Clone and cd into the repository.
--- a/syntheseus/reaction_prediction/environments/README.md
+++ b/syntheseus/reaction_prediction/environments/README.md
@@ -1,0 +1,28 @@
+# Single-step Model Environments
+
+Every single-step model may require a different environment and set of dependencies.
+Here we outline the steps to set up an environment for each of the supported models, which can be then used to run single-step model evaluation or multi-step search.
+
+## Basic setup
+
+All models apart from GLN can be set up using a shared base `conda` environment extended with a few model-specific dependencies. The general workflow is:
+
+```bash
+conda env create -f environment_shared.yml  # Create the shared environment.
+conda activate syntheseus-single-step       # Activate the environment.
+pip install -e ../../../                    # Install `syntheseus`.
+source setup_[MODEL_NAME].sh                # Run the extra setup commands.
+```
+
+If you wish to use several models, it's enough to create the environment once and run all the corresponding setup scripts.
+However, note that RetroKNN depends on LocalRetro, so if you want to use both, it is enough to run just `setup_retro_knn.sh`.
+
+The GLN model is not compatible with the others, currently requiring a specialized environment creation which includes building `rdkit` from source.
+We packaged all the necessary steps into a Docker environment defined in `gln/Dockerfile`.
+
+## Back-translation
+
+In `reaction_prediction/cli/eval.py` a forward model may be used for computing back-translation (round-trip) accuracy.
+Currently, Chemformer is the only supported forward model.
+
+To evaluate a particular model with back-translation computed using Chemformer, simply set up an environment for that model and then run `setup_chemformer.sh` on top.
--- a/syntheseus/reaction_prediction/environments/README.md
+++ b/syntheseus/reaction_prediction/environments/README.md
@@ -17,6 +17,9 @@ source setup_[MODEL_NAME].sh                # Run the extra setup commands.
 If you wish to use several models, it's enough to create the environment once and run all the corresponding setup scripts.
 However, note that RetroKNN depends on LocalRetro, so if you want to use both, it is enough to run just `setup_retro_knn.sh`.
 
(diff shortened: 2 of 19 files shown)


fiberleif (approved):
LGTM!

Just one question, l noticed that there is an `environment.yml` file in the root directory of Syntheseus. When should users install this file, and when should users install the `environment_shared.yml` file located in `reaction_prediction/environments directory` ? 

Some clarifications in the Readme might be helpful?

kmaziarz:
> l noticed that there is an `environment.yml` file in the root directory of Syntheseus. When should users install this file, and when should users install the `environment_shared.yml` file located in `reaction_prediction/environments directory` ?

When trying to set up just the core of `syntheseus` (to e.g. plug in your own single-step models) you'd use the environment from the top-level, while if trying to plug in specific single step models, you'd follow the instructions from `reaction_prediction/environments` (most of which start with using the `environment_shared.yml` file but not GLN).

We should think how/where to best explain this to the users though (perhaps easier when we also add the model checkpoints).

AustinT (approved):
Looks good to me. Only suggestion is updating top-level README.

syntheseus/reaction_prediction/environments/setup_shared.sh
```diff
@@ -5,7 +5,7 @@ mkdir -p external
 
 # Add the `external/` directory to `PYTHONPATH` when the environment is activated.
 mkdir -p $CONDA_PREFIX/etc/conda/activate.d
-echo "export PYTHONPATH=$PYTHONPATH:$PWD/external" >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
+echo "export PYTHONPATH=.:$PWD/external" >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
```
AustinT, line 8: Doesn't this assume the user's `PYTHONPATH` is empty?

kmaziarz: It ignores the pre-existing contents of `PYTHONPATH`. I thought that would make the setup more predictable, but maybe we shouldn't be so aggressive 😄 Although I guess we could prepend our `$PWD/external` (instead of appending) just in case the pre-existing `PYTHONPATH` contains directories named in the same way as one of our single-step models...

syntheseus/reaction_prediction/environments/gln/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+WORKDIR /
+
+# Install GLN (this relies on `CUDA_HOME` being set correctly).
+RUN git clone https://github.com/Hanjun-Dai/GLN.git
+WORKDIR /GLN
+RUN git checkout b5bd7b181a61a8289cc1d1a33825b2c417bed0ef
+RUN pip install -e .
+
+ENV PYTHONPATH=$PYTHONPATH:/rdkit:/GLN
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/rdkit/lib
```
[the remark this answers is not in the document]
kmaziarz: We can revisit cleaning up GLN's setup in the future... although last time I checked, I couldn't make it simpler, due to the requirement to build `rdkit` from source 😕

syntheseus/reaction_prediction/environments/setup_shared.sh
AustinT: I think any of these options are fine (including overwriting the PYTHONPATH), just wanted to ensure it wasn't an oversight.

syntheseus/reaction_prediction/environments/environment_shared.yml
```diff
@@ -0,0 +1,12 @@
...
+  - defaults
+  - conda-forge
+  - pytorch
+dependencies:
+  - numpy
+  - pandas
+  - pip
+  - python==3.9.7
+  - pytorch=1.10.2=py3.9_cuda11.3_cudnn8.2.0_0
+  - rdkit=2021.09.4
```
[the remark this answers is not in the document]
kmaziarz: The pin may be unnecessary; I'm rerunning the evaluation of all models to see if everything is fine when the CUDA version is unpinned.

README.md
```diff
@@ -6,17 +6,20 @@
...
+It contains implementations of common search algorithms, a simple API to wrap custom reaction models and write
+custom algorithms, and wrappers for many state-of-the-art reaction models from the literature.
 It is meant to allow for simple benchmarking of the components
 of retrosynthesis algorithms.
 
-## Installation
+## Setup
 
-Currently `syntheseus` is not hosted on PyPI
-(although this will likely change in the future).
```
kmaziarz, line 18: As the capabilities of the package expand, I think it's useful to keep the top-level README from growing too much. The remark about `syntheseus` not being on PyPI doesn't seem very useful, so I propose we drop it for conciseness.

syntheseus/reaction_prediction/environments/README.md
```diff
@@ -0,0 +1,28 @@
...
+conda activate syntheseus-single-step       # Activate the environment.
+pip install -e ../../../                    # Install `syntheseus`.
+source setup_[MODEL_NAME].sh                # Run the extra setup commands.
+```
+
+If you wish to use several models, it's enough to create the environment once and run all the corresponding setup scripts.
+However, note that RetroKNN depends on LocalRetro, so if you want to use both, it is enough to run just `setup_retro_knn.sh`.
+
+The GLN model is not compatible with the others, currently requiring a specialized environment creation which includes building `rdkit` from source.
+We packaged all the necessary steps into a Docker environment defined in `gln/Dockerfile`.
```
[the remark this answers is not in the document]
kmaziarz: Sounds good; see now?

README.md
AustinT: Good with me.

syntheseus/reaction_prediction/environments/README.md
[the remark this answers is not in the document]
AustinT: Looks good to me!

syntheseus/reaction_prediction/environments/environment_shared.yml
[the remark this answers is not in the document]
kmaziarz: Do you know how to unpin this in a way that works reliably? I tried truncating to `pytorch=1.10.2`, but by default that gets me the CPU-only version. I tried adding `cudatoolkit` to the list of dependencies, but when I start with a docker image that has CUDA 11.3, adding a `cudatoolkit` without a specific version gets me `cudatoolkit==11.8`. Even if I fix the version for `cudatoolkit` I still end up getting the CPU version of `pytorch`... 😕

(And if we cannot find a good solution, perhaps we can leave the explicit pin and make a note that this can be changed if desired?)

syntheseus/reaction_prediction/environments/setup_shared.sh
kmaziarz: I've now switched to prepending but keeping the current contents of `PYTHONPATH`.

PR #16, opened 2023-07-31
Change default node depth to infinity

AustinT:
*Background*: Previously a default value of `-1` was assigned to this value, mostly just because it was obviously not a valid depth so it would be clear to users that this is just the default value. However, in some recent experiments with graph algorithms I observed that this could occasionally cause a lot of unnecessary updates when adding a new node to a graph caused a cycle. The depth update uses the minimum depth of any parent, so sometimes the value of -1 was propagated around the graph for a while (giving its children a depth of 0, their children a depth of 1, etc) until this was eventually corrected and all the nodes were updated a second time. Even though each update was short, the cumulative effect was quite large for large graphs: in one experiment with 10k calls to the reaction model, 40min/60min was spent doing these depth updates.

*Rationale*: by changing the default to infinity these redundant updates do not occur because the depth update uses the minimum depth of any parent, so no node in the graph with finite depth will have its depth updated if a node is added whose depth value is set to infinite.

*Alternative considered*: instead of changing the default I could have done very careful updates to ensure that this propagation of -1 never happened (e.g. update the node depths in a correct order). However, changing the default is easier, and infinity is probably a more sensible default value for depth anyway since it would correspond to no valid path to the root node. This value would also be stable if depth updates were run on graphs without a root node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Change default node depth to infinity ([#16](https://github.com/microsoft/syntheseus/pull/16)) ([@austint])
 - Pin `pydantic` version to `1.*` ([#10](https://github.com/microsoft/syntheseus/pull/10)) ([@kmaziarz])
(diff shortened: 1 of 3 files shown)


kmaziarz (approved):
LGTM.

PR #17, opened 2023-08-02
Fix tutorials

jagarridotorres:
Adapted tutorials to the renaming from PR #9, as they used the old function name `min_cost_routes` instead of `iter_routes_cost_order`.

### DIFF
--- a/tutorials/search/1- end to end retrosynthesis with a toy model.ipynb
+++ b/tutorials/search/1- end to end retrosynthesis with a toy model.ipynb
@@ -472,7 +472,7 @@
     "from syntheseus.search.analysis import route_extraction\n",
     "\n",
     "# This iterator returns sets of nodes which constitute synthesis routes\n",
-    "for i, route_nodes in enumerate(iter_routes_cost_order.min_cost_routes(output_graph, max_routes=100)):\n",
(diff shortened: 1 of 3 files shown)


tutorials/search/1- end to end retrosynthesis with a toy model.ipynb
```diff
@@ -472,7 +472,7 @@
     "from syntheseus.search.analysis import route_extraction\n",
     "\n",
     "# This iterator returns sets of nodes which constitute synthesis routes\n",
-    "for i, route_nodes in enumerate(route_extraction.min_cost_routes(output_graph, max_routes=100)):\n",
+    "for i, route_nodes in enumerate(iter_routes_cost_order.min_cost_routes(output_graph, max_routes=100)):\n",
```
kmaziarz, line 475: This doesn't look right if you compare with the other call below?

kmaziarz (changes requested):

kmaziarz (approved):
Thanks for finding this! As discussed offline, we can also add a CHANGELOG entry and then this PR should be good to go.

AustinT (approved):
Looks good to me, thanks Jose!

jagarridotorres:
Thanks guys! Sorry for all those commits for such a small PR 😂

Btw, I really like the code!

PR #18, opened 2023-08-09
Add necessary node information for PDVN

fiberleif:
This PR is based on Austin's previous commit and adds necessary node information for PDVN.


AustinT (commented):
These changes seem like a reasonable addition to the PDVN PR, but shouldn't we maybe merge this into my PDVN PR branch rather than into main directly? This PR doesn't really make much sense on its own...

fiberleif:
> These changes seem like a reasonable addition to the PDVN PR, but shouldn't we maybe merge this into my PDVN PR branch rather than into main directly? This PR doesn't really make much sense on its own...

Hi, Austin, Yes, this PR is based on your previous one; however, I'm only covering the necessary node information in this small PR. We have successfully reimplemented PDVN based on your early PDVN PR and are currently verifying its performance. Once the performance is as good as the paper indicates, we will upload all of them in an organized manner. This PR is just to ensure the internal experiments run smoothly :)

PR #20, opened 2023-08-15
Fix environment setup script for MEGAN

kmaziarz:
There seem to be two issues with `setup_megan.sh` that evaded the testing done in #15:
- Two necessary dependencies (`gitpython` and `scipy`) are missing
- The version of `torchtext` used forces `pip` to switch to a different version of `torch` than the one used in the shared environment

This PR addresses both problems.


AustinT (approved):
I didn't run this, but it seems reasonable to me. Happy to approve a merge.

PR #21, opened 2023-08-15
Release single-step model checkpoints

kmaziarz:
Continuing from previous PRs (most notably #15), this PR releases trained model checkpoints for all currently integrated baselines.


AustinT (approved):
Nice table Kris. I tried all the links to check whether they are broken and it seems like they worked. However, I did not try running all the models. Do I understand correctly that they should all work with the existing `eval.py` script?

Also, I suggest making a CHANGELOG entry once #20 is merged.

fiberleif (approved):
LGTM!

kmaziarz:
> Do I understand correctly that they should all work with the existing eval.py script?

Yup, passed through the `model_dir` argument.

PR #22, opened 2023-08-25
Improve the packing number heuristic used for estimating diversity

kmaziarz:
To estimate the diversity of a set of routes, we currently use a heuristic packing number algorithm which finds the largest set of routes that are sufficiently different from each other (e.g. use non-overlapping reactions).

The current heuristic is _recursive_: to find the best diverse set, it divides the input list of routes into two halves, computes the result recursively, and then greedily combines the solutions. This is then repeated `num_tries` times and the best solution found is returned.

I compared this approach with two alternatives: a simpler _linear_ algorithm that reads the routes left-to-right and tries to greedily extend the solution, and a variant of the recursive algorithm that computes an _optimal_ way of combining the solutions from the recursive calls. To test the three approaches empirically, I took two search runs: a 5-minute run using Retro* + MEGAN, and a 15-minute run using MCTS + LocalRetro; the target molecules were selected so that many diverse routes are found. For each approach I varied the number of trials and tested it on several sets of routes extracted from the searches at various points in time.

In summary, the results are as follows: (1) the simple linear algorithm is faster than the current recursive one and on average finds slightly larger packing sets; (2) the improved recursive algorithm is the slowest per iteration but finds the best packing sets.

In this PR, I propose using the improved recursive algorithm for 10 trials instead of the current one for 100 trials. Empirically this is ~2x faster while finding slightly larger packing sets (~1 extra route in the best packing sets if the input is sufficiently challenging e.g. contains a packing set with 12 or more routes).

### DIFF
--- a/syntheseus/search/analysis/diversity.py
+++ b/syntheseus/search/analysis/diversity.py
@@ -40,8 +40,8 @@ def estimate_packing_number(
         distance_metric: distance between two routes.
         max_packing_number: to avoid expensive computations,
             the algorithm is stopped if a packing number
-            larger than this value is found.
-            If None, the algorithm will run until completion.
+            larger or equal to this value is found.
+            If `None`, the algorithm will run until completion.
         num_tries: the number of random restarts to perform.
         random_state: random state to use for shuffling routes.
 
@@ -115,7 +115,7 @@ def _recursive_construct_packing_set(
         return list(routes)
 
     # Recursive case:
-    # first calculate packing set for both halves
+    # First calculate packing set for both halves
     cutoff_idx = len(routes) // 2
     route_set1 = _recursive_construct_packing_set(
         routes[:cutoff_idx],
@@ -232,10 +232,10 @@ def molecule_symmetric_difference_distance(
     route2: SynthesisGraph,
 ) -> float:
     """
-    Calculate the symmetric difference distance between the sets of reactions in 2 routes.
+    Calculate the symmetric difference distance between the sets of molecules in 2 routes.
     """
 
-    # Get sets of reactions
+    # Get sets of molecules
     molecules1 = _get_molecules(route1)
     molecules2 = _get_molecules(route2)
     return len(molecules1 ^ molecules2)
--- a/syntheseus/search/analysis/diversity.py
+++ b/syntheseus/search/analysis/diversity.py
@@ -5,6 +5,8 @@ import random
 from collections.abc import Collection
 from typing import Callable, Optional
 
+import networkx as nx
+
 from syntheseus.search.chem import BackwardReaction, Molecule
 from syntheseus.search.graph.route import SynthesisGraph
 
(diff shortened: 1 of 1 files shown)


syntheseus/search/analysis/diversity.py
```diff
@@ -101,8 +103,8 @@ def _recursive_construct_packing_set(
     Recursive helper function for estimate_packing_number which finds a packing set.
 
     If <= 1 route is provided, the packing set is just the set of routes.
-    If >= 2 routes are provided, then the list is divided into two subsets
-    and the second subset is merged into the first,
+    If >= 2 routes are provided, then the list is divided into two subsets,
+    and the best solutions found within the subsets are merged,
     which requires at most (N/2)^2 distance computations.
```
AustinT, line 108:
```suggestion
    and a recursive call is made to find a packing set for each subset.
    These packing sets are then optimally merged, requiring at most (N/2)^2 distance computations.
    
    The details of the optimal merging are as follows:
    
    1. A graph is constructed between all routes in both packing sets A and B,
       with an edge between any two pairs of nodes with distance <= threshold.
       Because A and B are individually packing sets, this graph is *bipartite*,
       i.e. edges will only exist between nodes in A and nodes in B.
    2. A *minimum vertex cover* (a set of vertices where each edge in the graph
        links to at least one vertex in the set) is found by first finding a *maximum
        matching* ( a set of edges in a graph such that no two edges share a vertex),
        then applying Konig's theorem which guarantees a correspondence between
        a maximum matching and a minimum vertex cover for bipartite graphs.
        This is all implemented in networkx, which provides efficient algorithms.
    3. The nodes from the minimum vertex cover are removed from the graph.
        From the definition of vertex cover, this means that no edges will remain in the
        graph, and because a *minimum* vertex cover was found,
        this means that the smallest possible number of nodes was deleted.
        The remaining nodes now form a packing set.
```

syntheseus/search/analysis/diversity.py
```diff
@@ -130,35 +132,32 @@ def _recursive_construct_packing_set(
         max_packing_number,
     )
     assert (
-        max_packing_number is None or len(route_set1) <= max_packing_number
+        max_packing_number is None or max(len(route_set1), len(route_set2)) <= max_packing_number
     ), "Max packing number exceeded in recursive call."
```
AustinT, line 136: Would it make sense to add a case here to directly return `route_set1` or `route_set2` if one of these sets has size `max_packing_number`? That way we can avoid the expensive computation below.

AustinT: Suggestion:

```python
# If max packing number is reached, directly return one set without merging
if len(route_set1) == max_packing_number:
    return route_set1
elif len(route_set2) == max_packing_number:
    return route_set2
```

AustinT (approved):
Looks great Kris! I learned some interesting new graph theory. I just added a more detailed docstring justifying the technique, since it required a bit of thinking for me.

kmaziarz:
For posterity, here are the plots showing the comparisons I've mentioned in the PR description:

diversity_megan
diversity_local_retro

syntheseus/search/analysis/diversity.py
kmaziarz: Even better: we can check the size of `route_set1` after the first recursive call, and if it's already reaching max value, return without doing the other call (see my last commit). This could potentially give a massive speedup if `max_packing_number` is much smaller than the number of routes.

PR #23, opened 2023-08-30
Add node evaluators that use single-step model probability

kmaziarz:
In practical experiments, it is often useful to guide the search using the probabilities assigned by the single-step model, plugging them in either through a node value function or a search policy. This PR adds flexible node evaluators that enable this functionality.

The classes added here are covered by new tests in `node_evaluation/test_common.py`. However, graphs available as standard test fixtures in the codebase could not be used out-of-the-box as these graphs do not have single-step model probability metadata set. This is because they were produced using a `search`-style reaction model, whereas probability is only set by models wrapped with `SyntheseusBackwardReactionModel`. For now I adjusted the fixtures ad-hoc in the respective tests, but this discrepancy suggests that we might want to rethink how `search` and `reaction_prediction` interact.

### DIFF
--- a/syntheseus/search/node_evaluation/base.py
+++ b/syntheseus/search/node_evaluation/base.py
@@ -110,15 +110,15 @@ class ReactionModelBasedEvaluator(NoCacheNodeEvaluator[NodeType]):
                 `temperature != 1.0`.
             temperature: Temperature to apply (i.e. divide the logits by).
             clip_probability_min: Minimum probability to clip to. Should be positive if `return_log`
-                is set to avoid NaNs.
+                or `normalize` is set to avoid NaNs.
             clip_probability_max: Maximum probability to clip to.
         """
         super().__init__()
 
         assert 0.0 <= clip_probability_min <= clip_probability_max <= 1.0
 
-        if return_log and clip_probability_min == 0.0:
-            raise ValueError("Disabling clipping can lead to NaNs when computing log probability")
+        if (return_log or normalize) and clip_probability_min == 0.0:
+            raise ValueError("Disabling clipping can lead to NaNs")
 
         self._return_log = return_log
         self._normalize = normalize
--- a/syntheseus/search/node_evaluation/base.py
+++ b/syntheseus/search/node_evaluation/base.py
@@ -4,6 +4,9 @@ import abc
 from collections.abc import Sequence
 from typing import Generic, Optional, TypeVar
 
+import numpy as np
+
+from syntheseus.search.chem import BackwardReaction
 from syntheseus.search.graph.base_graph import RetrosynthesisSearchGraph
 from syntheseus.search.graph.node import BaseGraphNode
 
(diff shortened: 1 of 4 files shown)


syntheseus/search/node_evaluation/base.py
```diff
@@ -81,3 +84,53 @@ def _evaluate_nodes(
...
+        if "probability" not in metadata:
+            raise ValueError("Cannot call node evaluator as reaction model probability is not set")
+        return metadata["probability"]  # type: ignore
+
+    def _evaluate_nodes(self, nodes, graph=None) -> Sequence[float]:
+        probs = np.asarray([self._get_probability(n, graph) for n in nodes])
+        probs = np.clip(probs, a_min=self._clip_probability_min, a_max=self._clip_probability_max)
+
+        if self._return_log:
+            outputs = np.log(probs) / self._temperature
```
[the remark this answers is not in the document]
kmaziarz: Actually, the other branch of the `if` may need to be adjusted as well...

syntheseus/search/node_evaluation/base.py
```diff
@@ -81,3 +84,53 @@ def _evaluate_nodes(
     ) -> Sequence[float]:
         """Override this method to just evaluate the nodes, without counting the number of calls."""
         pass
+
+
+class ReactionModelBasedEvaluator(NoCacheNodeEvaluator[NodeType]):
+    """Evaluator that computes its value based on the probability from the single-step model."""
```
AustinT, line 90:
```suggestion
    """
    Evaluator that computes its value based on the `probability` metadata from the underlying reaction objects,
    with optional added re-normalization.
    """
```

syntheseus/search/node_evaluation/base.py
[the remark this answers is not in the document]
AustinT: I think we should either do nothing, or both divide by the temperature and softmax. Most likely retro* would be fine if the costs were normalized, but I think normalizing the costs doesn't really make sense for that algorithm.

syntheseus/search/node_evaluation/base.py
```diff
@@ -81,3 +84,53 @@ def _evaluate_nodes(
...
+
+        if self._return_log:
+            outputs = np.log(probs) / self._temperature
+        else:
+            outputs = probs ** (1.0 / self._temperature)
+
+        if self._return_negated:
+            outputs *= -1.0
+
+        return outputs.tolist()
```
AustinT, line 136: To be clear, the outputs are never normalized before being returned?

syntheseus/search/node_evaluation/common.py
```diff
@@ -15,3 +18,30 @@ def _evaluate_nodes(self, nodes, graph=None):
...
+
+class ReactionModelLogProbCost(ReactionModelBasedEvaluator[AndNode]):
+    def __init__(self, **kwargs) -> None:
+        super().__init__(return_log=True, return_negated=True, **kwargs)
+
+    def _get_reaction(self, node: AndNode, graph) -> BackwardReaction:
+        return node.reaction
+
+
+class ReactionModelProbPolicy(ReactionModelBasedEvaluator[MolSetNode]):
```
AustinT, line 31: Maybe add a docstring specifying that this is for OR-node MCTS?

AustinT (approved):
Looks good in general, although I think it would be good to add an option to normalized the probabilities (unless somehow it is already being done and I missed it).

AustinT:
Also, remember to add a CHANGELOG entry for this PR @kmaziarz

syntheseus/search/node_evaluation/base.py
kmaziarz: Yeah, I forgot about normalization; see last commit.

syntheseus/search/node_evaluation/base.py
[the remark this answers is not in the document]
kmaziarz: > I think we should either do nothing, or both divide by the temperature and softmax.

I'm not sure if I understand this, but see the last commit for what I've done regarding normalization - does this make sense?

I've set the default for `ReactionModelLogProbCost` to not normalizing, although I don't necessarily see if normalizing would make sense for Retro* or not. If `temperature = 1.0`, and the inputs are a valid probability distribution, normalization (as I implemented it now) will be a no-op. With `temperature != 1.0` and no normalization we will be essentially scaling all the costs by temperature. With `temperature != 1.0` + normalization something potentially non-trivial happens with the costs.

syntheseus/search/node_evaluation/base.py
```diff
@@ -93,10 +93,25 @@ def __init__(
...
     ) -> None:
+        """Initialized the evaluator.
+
+        Args:
+            return_log: Whether to return the logarithm of the probability instead of the
+                probability itself.
+            return_negated: Whether to multiply the result by `-1`.
+            normalize: Whether to renormalize the output to be a distribution (or logarithms of
+                probabilities corresponding to a valid distribution). This is mostly useful when
+                `temperature != 1.0`, as otherwise the inputs would typically be normalized.
```
AustinT, line 109: Not sure if I agree that they would "typically be normalized" since normalization in some models could either be over *all* reactions (but the model only returns topK), or the probability could be normalized over "whether or not the reaction should be included", i.e. not normalized w.r.t the other reactions.

syntheseus/search/node_evaluation/base.py
```diff
@@ -93,10 +93,25 @@ def __init__(
...
         temperature: float = 1.0,
         clip_probability_min: float = 1e-10,
         clip_probability_max: float = 0.999,
     ) -> None:
+        """Initialized the evaluator.
+
+        Args:
+            return_log: Whether to return the logarithm of the probability instead of the
+                probability itself.
+            return_negated: Whether to multiply the result by `-1`.
```
AustinT, line 106: Taking a second look at this: is there a reason we have a `return_negated` option instead of downstream classes just doing `-np.asarray(outputs).tolist()`?

syntheseus/search/node_evaluation/base.py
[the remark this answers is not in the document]
AustinT: Makes sense, I think the normalization commit resolves my concern here.

AustinT (approved):

syntheseus/search/node_evaluation/base.py
kmaziarz: Fair enough; I can drop the part after the comma then.

syntheseus/search/node_evaluation/base.py
kmaziarz: By "downstream classes", do you mean child classes of this, i.e. `ReactionModelLogProbCost`?

AustinT: Yes, exactly.

kmaziarz: Then there's no strong reason, negation could also happen in the child class. The only argument for having it in the base class would be if there were many classes applying this negation; then this would reduce some duplication as otherwise each of these subclasses would have to override `_evaluate_nodes` in the same way. But currently there is only one such class and maybe there won't be any others, so I can move the negation there.

PR #24, opened 2023-09-03
Fix error in graph standardization

AustinT:
The initial release included a file `syntheseus/search/graph/standardization.py` which contained code to turn all types of graphs into minimal AND/OR graphs (i.e. AND/OR graphs with one node per reaction and one node per molecule). 
It turns out that for MolSetGraph objects the code to fetch the reaction objects was incorrect and would crash. I didn't catch this before because I didn't actually write tests for this file.

This PR fixes this bug, and adds minimal "smoke" tests to make sure that this method runs without crashing. However, the actual output is not tested. To reflect this, the method still raises a warning when it is called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed bug where standardizing MolSetGraphs crashed ([#24](https://github.com/microsoft/syntheseus/pull/24)) ([@austint])
 - Change default node depth to infinity ([#16](https://github.com/microsoft/syntheseus/pull/16)) ([@austint])
 - Adapt tutorials to the renaming from PR #9 ([#17](https://github.com/microsoft/syntheseus/pull/17)) ([@jagarridotorres])
 - Pin `pydantic` version to `1.*` ([#10](https://github.com/microsoft/syntheseus/pull/10)) ([@kmaziarz])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- Fixed bug where standardizing MolSetGraphs crashed ([#24](https://github.com/microsoft/syntheseus/pull/24)) ([@austint])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed bug where standardizing MolSetGraphs crashed ([#24](https://github.com/microsoft/syntheseus/pull/24)) ([@austint])
```
kmaziarz, line 25: The [changelog format](https://common-changelog.org/#241-change) suggests using imperative mood for the entries (similarly to what's usually done for commit messages).
```suggestion
- Fix bug where standardizing MolSetGraphs crashed ([#24](https://github.com/microsoft/syntheseus/pull/24)) ([@austint])
```

syntheseus/tests/search/graph/test_standardization.py
```diff
@@ -1 +1,27 @@
...
+        output = get_unique_node_andor_graph(andor_graph_non_minimal)
+
+    assert len(output) == len(andor_graph_non_minimal)  # no nodes deleted here
+
+
+def test_smoke_molset(
+    molset_tree_non_minimal: MolSetGraph,
+):
+    with pytest.warns(UserWarning):
+        get_unique_node_andor_graph(molset_tree_non_minimal)
```
kmaziarz, line 27: Just noting that these function signatures would normally fit on a single line, and the reason why `black` doesn't reformat this is due to the [magic-trailing comma behaviour](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma) (it interprets the comma in the argument list as a strong preference against having the signature on a single line). I'm assuming it was not your intention to trigger this behaviour though 🙂
```suggestion
def test_smoke_andor(andor_graph_non_minimal: AndOrGraph):
    with pytest.warns(UserWarning):
        output = get_unique_node_andor_graph(andor_graph_non_minimal)

    assert len(output) == len(andor_graph_non_minimal)  # no nodes deleted here


def test_smoke_molset(molset_tree_non_minimal: MolSetGraph):
    with pytest.warns(UserWarning):
        get_unique_node_andor_graph(molset_tree_non_minimal)
```

kmaziarz (approved):
LGTM!

syntheseus/tests/search/graph/test_standardization.py
AustinT: Good point, this was not intentional. I just fixed it.

PR #25, opened 2023-09-03
Add a time limit multiplier to all algorithm base tests

AustinT:
Many algorithm tests involve running the algorithms for a fixed number of iterations or maximum time limit, then checking that a certain outcome happened. However, algorithms with a slow iteration speed may not achieve the desired outcome within the time limit if their iteration time is slow. I found an example of this when writing an algorithm that does a lot of computation with numpy arrays.

If we want slower algorithms to be testable by extending the `BaseAlgorithmTest`, I saw two options:

1. Extend the tolerances of all tests in the base class. This would have the side effect of making the tests for all algorithms take longer (e.g. 5s -> 30s)
2. Add a "time multiplier" which defaults to a small value, but can be overridden by child classes that want more time.

I decided that option 2 makes more sense, so this PR implements it. Specifically, I added an attribute `time_limit_multiplier` to `BaseAlgorithmTest`, and replaced all instances of `time_limit_s=X` with `time_limit_s=X*self.time_limit_multiplier` (except in the test for time limits themselves because there an absolute tolerance is tested). This can be overridden in child classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Added option to override time tolerance in algorithm tests ([#25](https://github.com/microsoft/syntheseus/pull/25)) ([@austint])
 - Improve the heuristic used for estimating diversity ([#22](https://github.com/microsoft/syntheseus/pull/22)) ([@kmaziarz])
 - Improve the aesthetics of `README.md` ([#19](https://github.com/microsoft/syntheseus/pull/19)) ([@kmaziarz])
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Added option to override time tolerance in algorithm tests ([#25](https://github.com/microsoft/syntheseus/pull/25)) ([@austint])
```
kmaziarz, line 12: Imperative mood (same comment as on #24)
```suggestion
- Add option to override time tolerance in algorithm tests ([#25](https://github.com/microsoft/syntheseus/pull/25)) ([@austint])
```

kmaziarz (approved):

PR #26, opened 2023-09-04
Add a top-level CLI for running end-to-end search

kmaziarz:
This PR adds a first version of a top-level CLI for running search. It supports all single-step models integrated so far, as well as both MCTS and Retro* as choices for the search algorithm. Both of these algorithms have reasonable defaults set; for now these don't depend on the single-step model, but after tuning I will add model-specific defaults.

As for the output from `search.py`, there are two considerations:
- On one hand, it's convenient for `search.py` to have options to extract and plot some routes right away, in order to be able to get search results by running a single script.
- On the other hand, after investigating the outputs, one may realize that while the search might have been successful, the extraction parameters should be adjusted, to e.g. avoid certain reaction steps or extract further routes. Thus, it is useful to be able to rerun the extraction without redoing the search.

To get best of both worlds, I propose for `search.py` to both extract and plot some routes (unless `num_routes_to_plot = 0`) as well as dump the full graph (for optional further analysis or re-extraction).

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import datetime
 import json
 import logging
+import math
 import pickle
+import statistics
 from dataclasses import dataclass, field
 from pathlib import Path
 from pprint import pformat
@@ -187,6 +189,8 @@ def run_from_config(config: SearchConfig) -> None:
     results_dir_current_run = results_dir_top_level / f"{config.model_class.name}_{str(timestamp)}"
 
     logger.info("Setup completed")
+
+    all_stats: List[Dict[str, Any]] = []
     for idx, smiles in enumerate(tqdm(search_targets)):
         logger.info(f"Running search for target {smiles}")
 
@@ -223,6 +227,7 @@ def run_from_config(config: SearchConfig) -> None:
             "soln_time_wallclock": soln_time_wallclock,
         }
 
+        all_stats.append(stats)
         logger.info(pformat(stats))
 
         with open(results_dir / "stats.json", "wt") as f_stats:
@@ -255,6 +260,29 @@ def run_from_config(config: SearchConfig) -> None:
                     visualize_andor(**visualize_kwargs)
                 else:
                     visualize_molset(**visualize_kwargs)
+
+        del results_dir
+
+    if len(search_targets) > 1:
+        logger.info(f"Writing summary statistics across all {len(search_targets)} targets")
+        combined_stats: Dict[str, float] = dict(
+            num_solved_targets=sum(stats["soln_time_wallclock"] != math.inf for stats in all_stats)
+        )
+
+        for key in [
+            "rxn_model_calls_used",
+            "num_nodes_in_final_tree",
+            "soln_time_rxn_model_calls",
+            "soln_time_wallclock",
+        ]:
+            values = [stats[key] for stats in all_stats]
+            combined_stats[f"average_{key}"] = statistics.mean(values)
+            combined_stats[f"median_{key}"] = statistics.median(values)
+
+        logger.info(pformat(combined_stats))
+
+        with open(results_dir_current_run / "stats.json", "wt") as f_combined_stats:
+            f_combined_stats.write(json.dumps(combined_stats, indent=2))
 
 
 def main(argv: Optional[List[str]]) -> None:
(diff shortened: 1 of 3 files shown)


syntheseus/cli/search.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if config.save_graph:
+            with open(results_dir / "graph.pkl", "wb") as f:
+                pickle.dump(output_graph, f)
+
+        if config.num_routes_to_plot > 0:
+            # Extract some synthesis routes in the order they were found
+            logger.info(f"Extracting up to {config.num_routes_to_plot} routes for analysis")
+
+            # TODO(kmaziarz): Add options to extract a diverse (or otherwise interesting) subset.
```
kmaziarz, line 238: In a future PR we can add `cli/analysis.py` and then have `cli/search.py` expose the options from `analysis.py` (but set the default analysis to something very lightweight, e.g. extracting a small diverse subset).

fiberleif (approved):
LGTM!

One minor question: l noticed that the search and saving of results are executed for each target molecule individually. Would it be good to 1) add some stats across target molecules here (e.g., average reaction model calls or time cost for the first solution), or 2) it should be written in another place to analyze the stats.json, which contains all the information for each target molecule?

kmaziarz:
> Would it be good to add some stats across target molecules here (e.g., average reaction model calls or time cost for the first solution)?

I think this doesn't hurt; I can add it.

syntheseus/search/utils/misc.py
```diff
@@ -0,0 +1,2 @@
+def lookup_by_name(module, name):
+    return module.__dict__[name]
```
AustinT, line 2: Nit: this is effectively just an alias for `module.__dict__`. Any reason you made this rather than just writing `module.__dict__`?

syntheseus/cli/search.py
```diff
@@ -0,0 +1,310 @@
...
+        build_node_evaluator("value_function")
+        build_node_evaluator("reward_function")
+        build_node_evaluator("policy")
+
+        alg_kwargs["bound_function"] = lookup_by_name(mcts_base, alg_kwargs["bound_function_class"])
+        del alg_kwargs["bound_function_class"]
+
+        alg = MolSetMCTS(**alg_kwargs)
+    else:
+        raise NotImplementedError(f"Unsupported search algorithm {config.search_algorithm}")
```
AustinT, line 198: This is a nice way of building the node evaluators!

syntheseus/cli/search.py
```diff
@@ -0,0 +1,310 @@
...
+                pickle.dump(output_graph, f_graph)
+
+        if config.num_routes_to_plot > 0:
+            # Extract some synthesis routes in the order they were found
+            logger.info(f"Extracting up to {config.num_routes_to_plot} routes for analysis")
+
+            # TODO(kmaziarz): Add options to extract a diverse (or otherwise interesting) subset.
+            routes: Iterator = iter_routes_time_order(
+                output_graph, max_routes=config.num_routes_to_plot
+            )
```
AustinT, line 262: Rather than iterate over minimum cost routes, we could also randomly sample some routes. I have some code for this in another repo that I could add later.

This is just a suggestion though, for now let's merge it like this.

syntheseus/cli/search.py
```diff
@@ -0,0 +1,310 @@
...
+                visualize_kwargs: Dict[str, Any] = dict(
+                    graph=output_graph,
+                    filename=str(results_dir / f"route_{route_idx}.pdf"),
+                    nodes=route,
+                )
+
+                if config.search_algorithm == "retro_star":
+                    visualize_andor(**visualize_kwargs)
+                else:
+                    visualize_molset(**visualize_kwargs)
```
AustinT, line 277:
```suggestion
                if isinstance(output_graph, AndOrGraph):
                    visualize_andor(**visualize_kwargs)
                elif isinstance(output_graph, MolSetGraph):
                    visualize_molset(**visualize_kwargs)
```

AustinT (approved):
Looks good to me!

syntheseus/search/utils/misc.py
kmaziarz: No strong reason; `module.__dict__` just looked a bit like a cryptic hack, and I thought it's maybe marginally nicer to have it wrapped in a utility function instead of repeating it...

PR #27, opened 2023-09-07
Handle the empty input case in ReactionModelBasedEvaluator

kmaziarz:
Node evaluators based on `ReactionModelBasedEvaluator` can perform normalization; if done in log space, this can fail if the input list is empty (due to taking `max` of an empty array). Whether empty lists are actually passed in for evaluation depends on the exact algorithm (it does not seem to happen in MCTS, but does happen in Retro*), so it's best to guard against that in the evaluator itself as done in this PR.


AustinT (approved):
Looks good to me.

However, I've seen the "empty list" edge case come up a few times myself. Perhaps a better long-term fix would be to implement this behaviour directly in the `__call__` method of the `NoCacheNodeEvaluator` base class?

PR #28, opened 2023-09-08
Speed up the diversity estimation heuristic

kmaziarz:
This PR applies three small tweaks to the diverse subset extraction heuristic, speeding it up without changing semantics:
- The recursive calls of `_recursive_construct_packing_set` all make use of the same (unchanging) list of routes, just consider different slices of that list. Instead of slicing the input list repeatedly in each recursive call (which does a lot of copying), it is more efficient to just pass in indices pointing to the slice under consideration (this changes the signature of the private `_recursive_construct_packing_set`, but the signature of the public `estimate_packing_number` remains unchanged).
- The `_jaccard_distance` function was instantiating both a set intersection and a set union of its arguments, whereas intersection size alone is enough to compute the distance.
- With `n <= 1` as a base case, `_recursive_construct_packing_set` was spending a lot of time on finding diverse subsets for inputs with tiny `n` e.g. `n = 2`, as there is overhead in e.g. creating tiny graphs. It is more efficient to make `n = 2` one of the base cases, as it's trivial to solve it optimally.

Empirically, these changes speed up the algorithm by 3%, 3% and 22%, respectively (together compounding to a speedup of around 26%).

### DIFF
--- a/syntheseus/search/analysis/diversity.py
+++ b/syntheseus/search/analysis/diversity.py
@@ -76,6 +76,8 @@ def estimate_packing_number(
 
         # Construct a packing set and check whether it is better than the previous one
         packing_set = _recursive_construct_packing_set(
+            0,
+            len(route_list),
             route_list,
             radius,
             distance_metric,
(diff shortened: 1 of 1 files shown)


AustinT (approved):
LGTM Kris! Thanks again for looking for efficiency improvements and doing careful speed tests!

It looks like you need to fix some pre-commit errors before it gets merged though.

PR #29, opened 2023-09-12
Fix error handling in MEGAN

kmaziarz:
In search experiments, it can be observed that MEGAN sometimes produces odd molecules containing `C+` atoms, which violate valence despite being accepted as valid by `rdkit`. These molecules fail sanitization through MEGAN's `fix_explicit_hs` utility, and so far our wrapper has been ignoring this and passing the molecule through to the model to propose reactions (likely pushing the problematic `C+` atom to reactants). In this PR, the underlying model is not called on inputs which fail `fix_explicit_hs`, and rather an empty list of reactions is returned in this case. Relatedly, I also silenced some of the other internal warnings produced by the MEGAN model by applying a (now improved) `suppress_outputs` context manager.

### DIFF
--- a/syntheseus/reaction_prediction/inference/megan.py
+++ b/syntheseus/reaction_prediction/inference/megan.py
@@ -9,7 +9,7 @@ The original MEGAN code is released under the MIT license.
 import os
 import sys
 from pathlib import Path
-from typing import List, Union
+from typing import Any, Dict, List, Optional, Union
 
 from rdkit import Chem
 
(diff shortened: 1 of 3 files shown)


syntheseus/reaction_prediction/inference/megan.py
```diff
@@ -9,7 +9,7 @@
 import os
 import sys
 from pathlib import Path
-from typing import List, Union
+from typing import Any, Dict, List, Optional, Union
```
AustinT, line 12: nit: would we not do `from __future__ import annotations` and use `dict`/`list`/etc? Or are you doing this for Pydantic?

AustinT (approved):
LGTM (although I only checked it superficially). It's nice to see a TODO comment in the code get dealt with!

syntheseus/reaction_prediction/inference/megan.py
kmaziarz: I can't recall what the issue with `pydantic` was anymore... Here it's mostly just old habits 🙂

PR #30, opened 2023-09-13
Set model-specific hyperparameter defaults for search

kmaziarz:
Following hyperparameter tuning on 25 targets from the ChemBL Hard set (as used in "Re-Evaluating Chemical Synthesis Planning Algorithms"), this PR places the best settings found in `cli/search_config.yml`, and makes `search.py` pick up the defaults from the file if one happens to be using a supported model/algorithm combination.

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -112,7 +112,7 @@ class SearchConfig(BackwardModelConfig):
     num_routes_to_plot: int = 5  # Number of routes to extract and plot for a quick check
 
 
-def run_from_config(config: SearchConfig) -> None:
+def run_from_config(config: SearchConfig) -> Path:
     set_random_seed(0)
 
     print("Running search with the following config:")
@@ -305,8 +305,10 @@ def run_from_config(config: SearchConfig) -> None:
         with open(results_dir_current_run / "stats.json", "wt") as f_combined_stats:
             f_combined_stats.write(json.dumps(combined_stats, indent=2))
 
+    return results_dir_current_run
 
-def main(argv: Optional[List[str]]) -> None:
+
+def main(argv: Optional[List[str]]) -> Path:
     config: SearchConfig = cli_get_config(argv=argv, config_cls=SearchConfig)
 
     def _warn_will_not_use_defaults(message: str) -> None:
@@ -345,7 +347,7 @@ def main(argv: Optional[List[str]]) -> None:
                     defaults={f"{config.search_algorithm}_config": relevant_defaults},
                 )
 
-    run_from_config(config)
+    return run_from_config(config)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


syntheseus/cli/search.py
```diff
@@ -304,10 +305,49 @@ def build_node_evaluator(key: str) -> None:
         with open(results_dir_current_run / "stats.json", "wt") as f_combined_stats:
             f_combined_stats.write(json.dumps(combined_stats, indent=2))
 
+    return results_dir_current_run
```
kmaziarz, line 308: As the exact directory is constructed on the spot (e.g. contains the timestamp), returning it is useful if one wants to call `search.py` from another script, e.g.
```python
results_dir = search.main(argv=None)

# Consume and analyze the results in `results_dir`.
with open(results_dir / "stats.json", "rt") as f_combined_stats:
    print(json.load(f_combined_stats))
    ...
```

AustinT (approved):
LGTM Kris!

One thought: I would be tempted to just manually set the clip probabilities (or at least the mins) instead of tuning them: these will probably only affect the search in extreme cases where the "best" reactions don't result in a solution, and it's not clear that reaction with tiny probabilities which are getting clipped _up_ to the minimum value are even realistic. That being said, reaction feasibility is a whole separate issue (as we acknowledge in the paper), so I am fine merging it as is. This is just a note for the future 🙂

kmaziarz:
> I would be tempted to just manually set the clip probabilities (or at least the mins) instead of tuning them: these will probably only affect the search in extreme cases where the "best" reactions don't result in a solution, and it's not clear that reaction with tiny probabilities which are getting clipped _up_ to the minimum value are even realistic. That being said, reaction feasibility is a whole separate issue.

Yeah, potentially discarding further outputs once a cumulative probability of e.g `0.995` is reached could help narrow down the top-`k` even more towards feasible reactions. That being said, without a way to assess feasibility, we can't check that making such a change actually helps, and the best cumulative probability threshold may also be different for different models.

I could potentially remove the clipping defaults (i.e. set them manually to a low min and a high max), rerun the Retro* Hard comparison, and if there is no significant difference (meaning that tuning these values didn't help) then keep it that way?

AustinT:
> I could potentially remove the clipping defaults (i.e. set them manually to a low min and a high max), rerun the Retro* Hard comparison, and if there is no significant difference (meaning that tuning these values didn't help) then keep it that way?

I think it's ok without doing this, but I am also happy for you to do it. Up to you!

kmaziarz:
> > I could potentially remove the clipping defaults (i.e. set them manually to a low min and a high max), rerun the Retro* Hard comparison, and if there is no significant difference (meaning that tuning these values didn't help) then keep it that way?
> 
> I think it's ok without doing this, but I am also happy for you to do it. Up to you!

(Having checked that the clipping doesn't seem to have a significant effect in most models, I have now removed those overrides, keeping them for MEGAN and RootAligned where clipping seemed to slightly improve results)

PR #31, opened 2023-10-26
Guard against rare issues in LocalRetro

kmaziarz:
In very rare cases LocalRetro's `get_k_predictions` can fail with `RuntimeError` originating from `rdkit`. I have not seen this with the USPTO-trained checkpoint, but it did happen once when evaluating a Pistachio-trained checkpoint (which has a much larger template diversity). This PR guards against this by simply falling back to returning no results if an error is raised.


AustinT (approved):
LGTM!

PR #32, opened 2023-11-08
Update the LocalRetro checkout to a newer commit

kmaziarz:
This PR updates the checkout of the external LocalRetro repository to pull in recent changes that happened upstream. This mostly enables training on larger datasets, but also brings a bit of improvement in dependency version flexibility and architecture customization.

Out of the box these new changes are not compatible with RetroKNN, which in one place depends on the bond order used by the underlying code, and that order has recently been changed. Concretely, RetroKNN contains code to convert the old bond order used by LocalRetro to a different one. As the latter coincides with the new order, this part of code can just be deleted.


funtion (approved):

AustinT (approved):
After a quick look (but not running the code) LGTM!

PR #33, opened 2023-11-09
Support *.csv and *.smi formats for the single-step evaluation data

kmaziarz:
Currently, running single-step evaluation requires supplying data in our special `*.jsonl` format. This means that a user needs to understand how the format is structured, and then convert their existing data. Instead of adding a conversion script, this PR extends the `DiskReactionDataset` class to support two additional formats: a csv-based one, commonly used to store raw USPTO-50K data, and a raw list-of-reaction-SMILES format. By default the right format is selected automatically based on the files present in the given directory. This allows running evaluation using the raw USPTO-50K data available on the web as-is. Moreover, converting from another format to the JSONL-based one is now just a matter of loading the data through the `DiskReactionDataset` class and then re-saving.


kmaziarz:
(Also, thanks to @panpiort8 for surfacing the input data format issue)

AustinT (approved):
I didn't run this / am not as close to the single-step code, but this seems reasonable to me. Happy to merge it if you feel it is a good improvement Kris.

PR #34, opened 2023-11-14
Add function to check if there is a route with a given set of starting molecules

AustinT:
In the paper [FusionRetro](https://proceedings.mlr.press/v202/liu23ah.html) they proposed checking whether the starting molecules for a route found during multi-step planning match the starting materials used in an actual synthesis route. This is essentially a less strict version of checking whether an exact ground-truth route is found. This PR implements code to check for the existence of a route with a specific set of starting materials.

In FusionRetro this metric is called "set-wise exact match", however this name is very vague so I instead called it `is_route_with_starting_mols`. I only wrote a version for AND/OR graphs because these typically contain a prohibitively large number of routes to explicitly enumerate them. The method is essentially a brute-force search with pruning of search branches. The inner workings of the methods are described reasonably-well in the docstrings (but I want to improve the description before merging)

I also fixed a random typo in a `conftest.py` docstring that I found while writing tests.

**TODO**:

- I need to add a CHANGELOG entry for this.
- should appr
- Check that pruned search works reasonably well in practice

### DIFF
--- a/syntheseus/search/analysis/starting_molecule_match.py
+++ b/syntheseus/search/analysis/starting_molecule_match.py
@@ -1,0 +1,62 @@
+"""Code related to starting molecules match metric (called exact set-wise match in FusionRetro)."""
+
+from __future__ import annotations
+
+import itertools
+import typing
+from collections.abc import Iterable
+
+T = typing.TypeVar("T")
+
+
+def partition_set(A: list[T], k: int) -> Iterable[list[list[T]]]:
+    """
+    Enumerates all possible ways to partition a list A into k disjoint subsets which are non-empty
+    (in all possible orders).
+
+    This function is easiest to explain by example:
+
+    A single partition is just the list A itself.
+
+    >>> list(partition_set([1, 2, 3], 1))
+    >>> [[[1, 2, 3]]]
+
+    For k=2, there are 3 possible partitions each with two possible orders.
+
+    >>> list(partition_set([1, 2, 3], 2))
+    >>> [[[1], [2, 3]], [[2], [1, 3]], [[3], [1, 2]], [[1, 2], [3]], [[1, 3], [2]], [[2, 3], [1]]]
+
+    For k=3, there is only 1 possible partition but 6 possible orders.
+
+    >>> list(partition_set([1, 2, 3], 3))
+    >>> [[[1], [2], [3]], [[1], [3], [2]], [[2], [1], [3]], [[2], [3], [1]], [[3], [1], [2]], [[3], [2], [1]]]
+
+    This function uses a recursive implementation.
+    First it partitions A into 2 (where the second partition has at least k-1 elements),
+    then it recursively partitions the second partition into (k-1) partitions.
+    """
+
+    # Check 1: elements of list are unique
+    assert len(set(A)) == len(A)
+
+    # Check 2: k is valid
+    assert k >= 1
+
+    # Base case 1: list is empty
+    if len(A) == 0:
+        return
+
+    # Base case 2: just a single partition
+    if k == 1:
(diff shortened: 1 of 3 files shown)


syntheseus/search/analysis/starting_molecule_match.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def _starting_mols_under_each_node(graph: AndOrGraph) -> dict[ANDOR_NODE, set[Molecule]]:
+    """Get set of molecules reachable under each node in the graph."""
+
+    # Initialize to empty sets, except for nodes with purchasable mols
+    node_to_mols: dict[ANDOR_NODE, set[Molecule]] = {n: set() for n in graph.nodes()}
+    for n in graph.nodes():
+        if isinstance(n, OrNode):
+            node_to_mols[n].add(n.mol)
```
kmaziarz, line 83: The comment says "purchasable", but is this reflected in the code?

kmaziarz (commented):
A very general question: what would happen if we just set the `is_purchasable` attribute to `False` for all nodes apart from those in `starting_mols`, and then computed the `has_solution` attribute using `run_message_passing`? Would that not give us what we want? Presumably this would give an affirmative answer even if a route exists that uses a _subset_ of `starting_mols`, so a related question would be whether we only care about routes that use exactly that set of leaves instead of its subset?

(Even more generally, I wonder how useful this graph-based formulation of this metric is: if a graph contains a combinatorial number of routes, then stating that there _exists_ a route with the right set of leaves seems not so practically useful. But I guess it doesn't hurt to have a function to compute this.)

AustinT:
> A very general question: what would happen if we just set the `is_purchasable` attribute to `False` for all nodes apart from those in `starting_mols`, and then computed the `has_solution` attribute using `run_message_passing`? Would that not give us what we want? Presumably this would give an affirmative answer even if a route exists that uses a _subset_ of `starting_mols`, so a related question would be whether we only care about routes that use exactly that set of leaves instead of its subset?
> 
> (Even more generally, I wonder how useful this graph-based formulation of this metric is: if a graph contains a combinatorial number of routes, then stating that there _exists_ a route with the right set of leaves seems not so practically useful. But I guess it doesn't hurt to have a function to compute this.)

You are right the `has_solution` approach would compute whether there exists a route that uses any *subset* of the starting molecules. The FusionRetro paper specifies an *exact* match. I'm not totally convinced by this metric but I think it is a proxy for a proposed route not being too different than a real route. I agree that the route existing "somewhere in the graph" is maybe not super useful.

I thought a bit about how to do this but could not think of a better way than what I coded: my impression is that a dynamic programming solution is not possible here because it there is no clear way to divide the problem into independent subproblems.

syntheseus/search/analysis/starting_molecule_match.py
AustinT: Good catch: I used to check whether a molecule is purchasable but removed it to make the method more general (i.e. the nodes marked purchasable in the graph don't need to be the starting molecules). I will fix this by changing the comment.

AustinT:
I converted this to a draft PR since some experiments I am running are very slow. I might change the implementation.

AustinT:
Just added some extra pruning code to increase the efficiency. I updated the main PR description to reflect this.

AustinT:
I think I will close this PR because:

1. I did not end up using the function it provided
2. In some instances the function was *very* slow (taking hours for just a modest graph). Unclear why this is.
3. I am not 100% sure it is correct (I wrote tests but they were not super thorough)
4. The precursor matching metric can be calculated directly from synthesis routes (added in #36)

We can re-visit and re-open this PR later if we want to add this function.

PR #35, opened 2023-11-15
Extend the single-step evaluation script with further metadata

kmaziarz:
This PR improves the reaction prediction evaluation script with two small extensions:
- attaching a `ground_truth_match` flag to the predictions to mark the ones that were judged as ground truth
- tracking the number of model predictions and reporting its simple statistics (for a variety of reasons the evaluated model may return less distinct outputs than was requested)

### DIFF
--- a/syntheseus/interface/models.py
+++ b/syntheseus/interface/models.py
@@ -31,7 +31,10 @@ class Prediction(GenericModel, Generic[InputType, OutputType]):
     score: Optional[float] = None  # Any other score.
     reaction: Optional[str] = None  # Reaction smiles.
     rxnid: Optional[int] = None  # Template id, if applicable.
-    metadata: Dict[str, Any] = {}  # Additional metadata.
+
+    # Dictionary to hold additional metadata. Note that we use a mutable default value here, which
+    # could be a problem in plain Python, but is handled correctly by `pydantic`.
+    metadata: Dict[str, Any] = {}
 
     def get_prob(self) -> float:
         if self.probability is not None:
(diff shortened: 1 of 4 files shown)


syntheseus/interface/models.py
```diff
@@ -31,7 +31,7 @@ class Config:
     score: Optional[float] = None  # Any other score.
     reaction: Optional[str] = None  # Reaction smiles.
     rxnid: Optional[int] = None  # Template id, if applicable.
-    metadata: Optional[Dict[str, Any]] = {}  # Additional metadata.
+    metadata: Dict[str, Any] = {}  # Additional metadata.
```
AustinT, line 34: Are you sure that setting the default to `{}` here avoids creating a shared dict for all instances of this class? I recall that python does this often, and this is why for example people avoid using dicts/lists as default kwargs and why `dataclass` uses `default_factory`. I recall `default_factory` being an issue for `pydantic` but can't remember precisely how this was resolved. Just thought I should flag this as a potential issue.

syntheseus/interface/models.py
```diff
@@ -67,7 +67,7 @@ class Config:
 
     input: InputType
     predictions: List[Prediction[InputType, OutputType]]
-    metadata: Optional[Dict[str, Any]] = {}
+    metadata: Dict[str, Any] = {}
```
AustinT, line 70: Same comment as above.

syntheseus/reaction_prediction/cli/eval.py
```diff
@@ -22,6 +22,7 @@
 from enum import Enum
 from functools import partial
 from itertools import islice
+from statistics import mean, median
```
AustinT, line 25: ... I never knew there was a statistics module in the standard library- cool!

AustinT (approved):
LGTM!

syntheseus/interface/models.py
AustinT: Great investigation Kris! Do you think it would be worth putting a small comment to note this behaviour? It could save us some trouble if we ever switch away from pydantic in the future.

kmaziarz: Make sense, done now.

PR #36, opened 2023-11-16
Add method to get starting molecules from synthesis route.

AustinT:
This is useful for various analysis techniques, e.g. the "starting molecule matching" metric from FusionRetro.

**TODO**: need to add CHANGELOG entry.

### DIFF
--- a/syntheseus/search/graph/route.py
+++ b/syntheseus/search/graph/route.py
@@ -61,5 +61,18 @@ class SynthesisGraph(BaseReactionGraph[BackwardReaction]):
     ) -> Sequence[BackwardReaction]:
         raise NotImplementedError
 
+    def get_starting_molecules(self) -> set[Molecule]:
+        """
+        Get the 'starting molecules' for this route,
+        i.e. reactant molecules which are not a product of a child reaction.
+        """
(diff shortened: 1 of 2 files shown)


syntheseus/search/graph/route.py
```diff
@@ -61,5 +61,18 @@ def expand_with_reactions(
...
         raise NotImplementedError
 
+    def get_starting_molecules(self) -> set[Molecule]:
+        """
+        Get the 'starting molecules' for this route,
+        i.e. reactant molecules which are not a product of a child reaction.
+        """
+        output: set[Molecule] = set()
+        for rxn in self._graph.nodes:
+            successor_products = {rxn.product for rxn in self.successors(rxn)}
```
kmaziarz, line 71: So `rxn` from the outer scope is passed as argument to `successors` and then shadowed by another variable called `rxn`? I guess the behaviour is correct but perhaps can be a bit confusing to read?

kmaziarz (approved):
LGTM.

syntheseus/search/graph/route.py
AustinT: Fixed!

PR #39, opened 2023-11-21
Fix some typos and unclear messages

AustinT:
First fix: error message for converting an AND/OR graph to a synthesis route would say there was more than one node if in fact there were zero nodes.

Second fix: reaction string in `conftest.py`. This was also fixed in #34, but since I am not merging that I decided to fix it here.


kmaziarz (approved):
LGTM.

PR #42, opened 2023-12-04
Add a simple CI for integration testing

kmaziarz:
Following simplifications to single-step model setup in #41, this PR adds a minimalistic integration test, which for now just checks that doing `pip install .[all]` works. In the future this will be extended by running minimal single-step model inference and multi-step search using the supported models.

### DIFF
--- a/.github/workflows/ci_single_step.yml
+++ b/.github/workflows/ci_single_step.yml
@@ -1,0 +1,27 @@
+name: CI (integration tests)
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
+    name: build
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


AustinT (approved):
LGTM. Only note is that by installing `environment_full` we might miss errors in the dependencies specified in `pyproject.toml` (e.g. the pandas dependency in MHN). However, I don't think this concern is sufficiently pressing to not merge this PR; it's just something to keep in mind for the future.

kmaziarz:
> LGTM. Only note is that by installing `environment_full` we might miss errors in the dependencies specified in `pyproject.toml` (e.g. the pandas dependency in MHN).

But this would be detected by the pre-existing CI, which uses the minimalistic environment (which for some reason still installs `pandas`, but anyway...)?

AustinT:
> > LGTM. Only note is that by installing `environment_full` we might miss errors in the dependencies specified in `pyproject.toml` (e.g. the pandas dependency in MHN).
> 
> But this would be detected by the pre-existing CI, which uses the minimalistic environment (which for some reason still installs `pandas`, but anyway...)?

Good point, I forgot about that. Looks good then!

PR #45, opened 2023-12-11
Add initial documentation

kmaziarz:
This PR adds an initial static documentation built using [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/). The docs are built automatically using GitHub Actions and deployed to [microsoft.github.io/syntheseus/](https://microsoft.github.io/syntheseus/).

This first version is a combination of content we had in `README.md` (which is changed here to be very minimalistic, linking to the docs for further information) and the pre-existing `docs/cli/eval_single_step.md`. In future PRs we can extend the docs with documentation for the search CLI, as well as an API reference and tutorials.


kmaziarz:
Note that rebuilding of the docs happens either when a PR is merged into `main` or when one triggers it manually from the Actions tab. This means that, when testing a PR that changes the docs, running the action would also publish the changes to be externally visible _before_ the PR is actually merged (although this would only be the case for the final test, as `mkdocs` can serve the work-in-progress locally; the result of running the action can also be easily reverted by re-running the action from `main`). This setup seems fine for now, although if we want we could consider enabling some sort of versioning in the future (likely there is a good solution provided by Materials for MkDocs, I just haven't looked into that yet).

AustinT (approved):
Looks fantastic Kris, thanks for pushing this forward!

PR #46, opened 2023-12-12
Add PDVN into the search CLI

fiberleif:
This PR adds the PDVN search algorithm to `cli/search`, and the `search_algorithm` argument now supports `retro_star`, `mcts`, and `pdvn`.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 ### Added
 
 - Add a general CLI endpoint ([#44](https://github.com/microsoft/syntheseus/pull/44)) ([@kmaziarz])
+- Add PDVN search algorithm into cli/search ([#46](https://github.com/microsoft/syntheseus/pull/46)) ([@fiberleif])
```
kmaziarz, line 13:
```suggestion
- Add support for PDVN to the search CLI ([#46](https://github.com/microsoft/syntheseus/pull/46)) ([@fiberleif])
```

syntheseus/cli/search.py
```diff
@@ -1,7 +1,7 @@
 """Script for running end-to-end retrosynthetic search.
 
-The supported single-step model types are listed in `syntheseus/reaction_prediction/cli/eval.py`;
-each can be combined with either MCTS or Retro* to perform search.
+The supported single-step model types are listed in `syntheseus/cli/eval_single_step.py`;
```
kmaziarz, line 3: Good catch!

syntheseus/cli/search.py
```diff
@@ -79,6 +80,27 @@ class MCTSConfig:
...
+    value_function_syn_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.5})
+
+    value_function_cost_class: str = "ConstantNodeEvaluator"
+    value_function_cost_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.0})
+
+    and_node_cost_fn_class: str = "ConstantNodeEvaluator"
+    and_node_cost_fn_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.1})
+
+    policy_class: str = "ReactionModelProbPolicy"
+    policy_kwargs: Dict[str, Any] = field(default_factory=dict)
```
kmaziarz, line 97: Is it then the case that it's not yet possible to plug in a PDVN-improved policy here (and so for now PDVN would be expected to behave similarly to MCTS)?

kmaziarz: OK, so essentially the version we integrate here does not allow one to reproduce the results from the original PDVN paper, as that would require plugging in _two_ models (i.e. a separate policy model different from the reaction model), which the current integration does not allow, right? But this is fine as a stepping-stone towards full integration.

kmaziarz (approved):

syntheseus/cli/search.py
fiberleif: Yeah, it's true

PR #47, opened 2023-12-13
Return model predictions as dataclasses instead of pydantic models

kmaziarz:
The `Prediction` and `PredictionList` objects were made based on `pydantic`, anticipating this will make it easier to e.g. expose a model through an API using a framework such as FastAPI. However, this meant introducing `pydantic` into the core of the library in order to facilitate a feature that most users may not care about. On the other hand, FastAPI can deal with non-`pydantic` classes too (`pydantic` is only needed for some extras like validation). Thus, once we want to integrate the server code properly, we should do it in such a way that `pydantic` (and further dependencies like `fastapi`) remain optional extras rather than part of the core of the library.

For now, in this PR I make the core prediction objects plain dataclasses, and drop `pydantic` from dependencies. This required a few changes (e.g. dropping special handling of prediction classes in `dictify`, using `default_factory` for the dict-valued fields), as well as fixing some of the tests, which did not provide all fields needed to construct the prediction objects (which for some reason is still accepted by `pydantic`).

### DIFF
--- a/syntheseus/interface/models.py
+++ b/syntheseus/interface/models.py
@@ -2,10 +2,8 @@ from __future__ import annotations
 
 import math
 from abc import abstractmethod
+from dataclasses import dataclass, field
 from typing import Any, Dict, Generic, List, Optional, TypeVar
-
-from pydantic import root_validator
-from pydantic.generics import GenericModel
 
 from syntheseus.interface.bag import Bag
 from syntheseus.interface.molecule import Molecule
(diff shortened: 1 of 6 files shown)


syntheseus/interface/models.py
```diff
@@ -32,9 +27,14 @@ class Config:
...
-    # Dictionary to hold additional metadata. Note that we use a mutable default value here, which
-    # could be a problem in plain Python, but is handled correctly by `pydantic`.
-    metadata: Dict[str, Any] = {}
+    # Dictionary to hold additional metadata.
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+    def __post_init__(self):
+        if self.probability is not None and self.log_prob is not None:
+            raise ValueError(
+                "Probability can be stored as probability or log probability, not both"
```
mrwnmsr, line 36: why is that?

mrwnmsr (approved):

AustinT (approved):
LGTM!

PR #48, opened 2023-12-14
Automatically download a default checkpoint if it is not provided

kmaziarz:
This PR extends the single-step model classes to automatically download the default (trained on USPTO-50K) model checkpoint if no model directory is provided. This makes it even easier to get started with `syntheseus` as one does not have to manually download and unzip the files. Once downloaded, the checkpoints are cached.

On top of this, I also extend the integration CI to run a new single-step model test that checks that the supported models do run and return reasonable predictions for a simple input. While locally the test works for all 6 `pip`-installable models, on GitHub the GPU is not available, and setting `device="cpu"` uncovers that some of the models appear to only work on GPU. Thus, for now the test only covers Chemformer, LocalRetro and MEGAN; the other models will be added in the future.

When running those tests, I found that the links we provided for LocalRetro and RetroKNN were not quite correct (the initial upload missed some of the files, which I later corrected, but the links were still pointing to the first upload rather than the corrected one). Morever, I found that RetroKNN was zipped using a method that `python` could not deal with, so I re-generated the file and re-uploaded it (again fixing the link).

The changes are separated into reasonable commits for easier reviewing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 - Add a general CLI endpoint ([#44](https://github.com/microsoft/syntheseus/pull/44)) ([@kmaziarz])
 - Add support for PDVN to the search CLI ([#46](https://github.com/microsoft/syntheseus/pull/46)) ([@fiberleif])
+- Add initial static documentation ([#45](https://github.com/microsoft/syntheseus/pull/45)) ([@kmaziarz])
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 ### Changed
 
 - Select search hyperparameters depending on which algorithm and single-step model are used ([#30](https://github.com/microsoft/syntheseus/pull/30)) ([@kmaziarz])
-- Add option to override time tolerance in algorithm tests ([#25](https://github.com/microsoft/syntheseus/pull/25)) ([@austint])
 - Improve the heuristic used for estimating diversity ([#22](https://github.com/microsoft/syntheseus/pull/22), [#28](https://github.com/microsoft/syntheseus/pull/28)) ([@kmaziarz])
-- Improve the aesthetics of `README.md` ([#19](https://github.com/microsoft/syntheseus/pull/19)) ([@kmaziarz])
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Changed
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -23,9 +23,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 ### Changed
 
 - Select search hyperparameters depending on which algorithm and single-step model are used ([#30](https://github.com/microsoft/syntheseus/pull/30)) ([@kmaziarz])
-- Add option to override time tolerance in algorithm tests ([#25](https://github.com/microsoft/syntheseus/pull/25)) ([@austint])
```
AustinT, line 26: Why is this line deleted? I don't think this was removed.

CHANGELOG.md
```diff
@@ -23,9 +23,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
...
 - Improve the heuristic used for estimating diversity ([#22](https://github.com/microsoft/syntheseus/pull/22), [#28](https://github.com/microsoft/syntheseus/pull/28)) ([@kmaziarz])
-- Improve the aesthetics of `README.md` ([#19](https://github.com/microsoft/syntheseus/pull/19)) ([@kmaziarz])
```
AustinT, line 28: Why is this line deleted? You don't change the README in this PR.

syntheseus/reaction_prediction/utils/downloading.py
```diff
@@ -0,0 +1,74 @@
...
+    if not any(cache_dir.iterdir()):
+        cache_zip_path = cache_dir / "model.zip"
+
+        logger.info(f"Downloading data from {link} to {cache_zip_path}")
+        urllib.request.urlretrieve(link, cache_zip_path)
+
+        with zipfile.ZipFile(cache_zip_path, "r") as f_zip:
+            f_zip.extractall(cache_dir)
+
+        cache_zip_path.unlink()
```
AustinT, line 44: Is there any reason why you don't use the `tempfile` package for this? Is it because you want to save the file between runs?

AustinT (approved):
Overall LGTM, but I just skimmed through the code rather than actually running it, so it is very possible that I didn't catch everything.

syntheseus/reaction_prediction/utils/downloading.py
kmaziarz: Yes, exactly: downloading and unzipping can take at least a few seconds, and in some cases (e.g. RetroKNN with a 3GB checkpoint) several minutes. I wanted to make sure we only do this once. This also allows us to add more tests for the models into the integration CI pipeline without incurring this cost for every test.

CHANGELOG.md
kmaziarz: (As discussed offline, I sneaked in a slight cleanup of unimportant CHANGELOG entries)

CHANGELOG.md
kmaziarz: (As discussed offline, I sneaked in a slight cleanup of unimportant CHANGELOG entries)

PR #50, opened 2023-12-18
Make the package compatible with PyPI

kmaziarz:
This PR makes the `syntheseus` package work correctly on PyPI, which required several small changes:
- Using an absolute link for the logo, as otherwise it won't render in the PyPI package description.
- Adding project links to give PyPI access to GitHub's metadata.
- Switching to getting the single-step model code from auxiliary PyPI uploads instead of GitHub (PyPI doesn't allow the latter).
- Removing author email addresses, as PyPI cannot handle them in the case of multiple authors (and, while I was at it, also adding @fiberleif).


fiberleif (approved):
LGTM!

PR #53, opened 2024-01-03
Make single-step models less noisy

kmaziarz:
This PR combines a sequence of changes to make the single-step models less noisy. This involves silencing benign warnings originating from PyTorch, turning off logging produced by tensorflow and OpenNMT, and also making sure the underlying model code doesn't tamper with root logger handlers or create log files. With these changes, creating and using any of the models should be now free of confusing logging/warning messages.


AustinT (approved):
LGTM, these all seem reasonable! Overwriting environment variables for MEGAN is a bit hacky, but imo not too hacky to merge.

fiberleif (approved):

PR #56, opened 2024-01-08
Add flags to control canonicalization during data loading

fiberleif:
This PR adds the options of `canonicalize` and `ordered` to the basic data classes and relevant functions, such as `DiskReactionDataset`, `Bag`, and `undictify_bag_of_molecules`. It provides flexibility and allows the reactants/products to be stored in their original SMILES, which is essential for training SMILES-based Transformer models.

 The default values of `canonicalize` and `ordered` have been set to `True` and `False` respectively, to avoid affecting existing code. These options can be opened when needed (e.g., when initializing a `DiskReactionDataset` object for training).


syntheseus/interface/bag.py
```diff
@@ -19,8 +19,11 @@ class Bag(Collection, Generic[ElementT]):
     linear in terms of the size of the bag.
     """
 
-    def __init__(self, values: Iterable[ElementT]) -> None:
-        self._values = tuple(sorted(values))
+    def __init__(self, values: Iterable[ElementT], ordered: bool = False) -> None:
+        if ordered:
+            self._values = tuple(values)
+        else:
+            self._values = tuple(sorted(values))
```
AustinT, line 26: If we allow the elements of the `Bag` to be ordered, does that not break its interpretation as a frozen multi-set? If order matters, why not just use a `tuple`?

AustinT (commented):
I think adding optionally canonicalization could makes sense, although it does create the possibility for multiple non-equal versions of molecules to be created which could cause problems downstream (e.g. checking whether reactants equal products). Is it really "essential" to train transformers with non-canonical SMILES?

I also don't really understand the modifications you made to the `Bag` data structure. Allowing ordering breaks its interpretation as a frozen multi-set. In these cases would a `tuple` not be a more suitable data structure? When would an ordered `Bag` be useful?

kmaziarz:
> I think adding optionally canonicalization could makes sense, although it does create the possibility for multiple non-equal versions of molecules to be created which could cause problems downstream (e.g. checking whether reactants equal products). Is it really "essential" to train transformers with non-canonical SMILES?

Good point: while the `canonicalize` flag already exists in the codebase, I think so far it has been used to signal that the caller _knows_ the SMILES to already be canonical, and so canonicalization can be skipped; we still assumed all `Molecule` objects always contain canonical SMILES. The usecase here is different, as `Molecule`s with non-canonical SMILES would be created "on purpose" (as part of data augmentation). I think this may be necessary for good performance in SMILES-based transformers, although I'm not sure if this should warrant changes to `syntheseus`, given that currently `syntheseus` does not handle any model training, only inference...

@fiberleif: Maybe we should actually first take a look at the transformer model training code, and see to what extent it needs to be tied to `syntheseus` and it's internal datastructures (`Molecule` / `Bag`)?

syntheseus/interface/bag.py
kmaziarz: I think these changes came from wanting to use the `DiskReactionDataset` class for loading data for training SMILES LMs. We may want to take a look at how much we're actually gaining from doing this.

AustinT:
> The usecase here is different, as `Molecule`s with non-canonical SMILES would be created "on purpose" (as part of data augmentation). I think this may be necessary for good performance in SMILES-based transformers, although I'm not sure if this should warrant changes to `syntheseus`, given that currently `syntheseus` does not handle any model training, only inference...

Thinking a little bit more, my suggestion for how to store augmented non-canonical SMILES would be in a molecule's `metadata` field. This does not affect comparison and can be used to store arbitrary things.

kmaziarz:
Closing this as we refactored the internal code so that these changes are no longer needed.

PR #57, opened 2024-01-10
Get all models to work on CPU

kmaziarz:
PR #48 added a new CI pipeline that covers single-step model inference, but it only tested half of the single-step models as the other ones did not work on CPU. This PR fixes this, making all models work on CPU, which mostly required making sure the right device is correctly propagated into all of the model internals. I then extended the test which now checks all `pip`-installable models we support.

Additionally, I also merged the two CI pipelines into one, as this allows for generating a combined coverage report that takes into account all tests. This does not affect testing speed, as the different test jobs still run in parallel. Finally, I also noticed that the `.coveragerc` configuration was not quite right, as the files with 0% coverage were not being discovered (e.g. see [this CI run on `main`](https://github.com/microsoft/syntheseus/actions/runs/7455113293/job/20283663956) which did not report the 0% coverage of `cli/search.py`); after a minor change to the configuration this is now fixed (see [here](https://github.com/microsoft/syntheseus/actions/runs/7468806136/job/20324898126)).

The changes are separated into self-contained commits for easier reviewing.


.github/workflows/ci.yml
```diff
@@ -33,6 +33,60 @@ jobs:
...
       run: |
-        python -m pytest --cov --cov-config=.coveragerc ./syntheseus/tests/
+        coverage run -p -m pytest ./syntheseus/tests/
+        coverage report --data-file .coverage.*
+    - name: Upload coverage report
+      uses: actions/upload-artifact@v4
+      with:
+        name: .coverage.core-${{ matrix.python-version }}
+        path: .coverage.*
+  test-models:
```
AustinT, line 45: Are we running the single-step model tests separately to avoid re-running with multiple python versions (to save compute), or because they only work with a very specific python version?

AustinT (approved):
LGTM. The improvements to the CI checks are cool (I didn't know you could upload/download artifacts and use them in later checks). Thanks for fixing coverage 😃

.github/workflows/ci.yml
kmaziarz: The single step models tests use full a syntheseus installation, whereas the core tests use the minimal installation. By separating the two we get to test both setups. But yes, it's also true that the single-step models have pinned dependencies and we would need to define separate pins to run them under Python 3.7 (although they may actually run, I haven't checked).

AustinT: Makes sense. Let's just keep it like this then. Creating a full python 3.7 environment with all dependencies does not sound like a useful exercise to me (especially since 3.7 is deprecated now).

PR #58, opened 2024-01-10
Potential merger of two reaction model classes

AustinT:
We agreed that having two interfaces for essentially the same thing is a bit awkward. This PR is a proof-of-concept merger of the separate reaction models from the `search` and `reaction_prediction`. It proposes reasonable compromise classes and shows that all the tests can pass with minimal modification if the classes are merged.

> [!IMPORTANT]
> There are a few issues that would need to be worked out more fully before merging this PR into main, so treat this PR more as a starting point for discussion than changes which should actually be merged in their current form. _Therefore this is just a draft PR_

# Change 1: merging reaction classes

## Summary of changes

The two kinds of reaction models use different classes: `syntheseus.search.chem.BackwardReaction` and `syntheseus.interface.models.Prediction` (and its subclasses). I merged them into `syntheseus.interface.models.Prediction`, keeping the features marked in **bold**:

| `search` | `reaction_prediction` |
| --- | --- |
| names are `reactants/product` | **names are `input/output`** |
|Reactants are sets|reactants are **bags**|
|Metadata is **typed dict**|metadata is plain dict|
|**has optional identifier field (defaults to None)**|no identifier field|
|scores/probs/etc are in metadata (**and not hashed/compared**)|**explicit fields for score/prob/etc** (which are hashed/compared)|

As an interim measure, I also augmented the `BackwardPrediction` a subclass that aliases `input/output` with `product/reactants`.

## Thoughts on changes

Overall this was a fairly simple merger.

From the search side, the only complication is that reactants were previously assumed not to contain duplicates, but I fixed this by replacing `for x in r.reactants` with `for x in set(r.reactants)` in a few places. I could see this potentially causing errors in the future if users are not careful though. If this is a concern, we could create a `unique_reactants` alias (or something similar).

From the reaction prediction side, the only thing lost is that two identical reactions with different scores or probabilities are now considered equal. However, my understanding is that this functionality was not used so it shouldn't matter.

Although this class was a workable compromise, I would also advocate for the following changes (which are not implemented in this PR):

1. Move score/probability/etc fields to `metadata`. They are not used by all models, and the purpose of having a `metadata` dict was specifically to avoid every reaction object carrying around a bunch of unused attributes filled with `None`.
2. I think the class should probably called `Reaction` instead of `Prediction`. This name is clearer, and I imagine syntheseus users would expect us to have a Reaction class and find the name "Prediction" unintuitive. At the same time, if this class will be used throughout syntheseus, I think the name "Prediction" is too broad, since syntheseus can be used to predict other things (e.g. entire synthesis routes)

# Change 2: merging reaction model classes

## Summary of changes

I merged `syntheseus.search.reaction_models.BackwardReactionModel` into `syntheseus.search.interface.models.ReactionModel` class. These classes had more differences, which are in the table below. Features which made it into the merged class are in **bold**.

| `search` | `interface` (old) |
| --- | --- |
| **tracks number of calls**|no tracking|
|**caches output** (1)|no caching|
|**configurable duplicate removal (true by default)** (2)|None|
|Returns `list[list[reaction]]`|**returns PredictionList** (3)|
|no model info|**include `get_model_info` and `get_parameters`**|
|Call signature is `list[mol]`|Call signature is **`list[mol], num_results`** (4)|
|**Subclasses override `_get...` method** | subclasses override `__call__` directly (5)|

Elaboration on above:

1. To handle `num_results`, the cache keys are now `(mol, num_results)`.
2. More on this below. I mainly didn't want to change all the tests for search.
3. There were some typing issues here since I subclasses `BackwardPrediction` and added some extra methods, so we might need to change the types a bit if we keep this change. It runs fine though.
4. To handle calls in search I made `num_results` optional. If not provided it is filled with a model's `default_num_results` attribute, which I set to 100 by default (let me know if it should be even higher).
5. This was done to ensure that caching and call counting is done.

## Thoughts on changes

Overall this merger had a few pain points but was easier than expected. That being said, there are a few awkward issues:

- Separately caching each number of results is inefficient if the output with `num_results=N` will always match the first `N` entries of the output with `num_results=(N+M)`. Overall this is probably minor though.
- It is unclear to me whether caching will mess up the evaluation script in any way (e.g. via timing). One solution would just be to turn off caching in this script.
- Similarly, it is unclear whether duplicate removal will mess up the eval script (but we could just turn it off in that script and call `model.reset()`). We could avoid this by setting it to `False` by default, but besides for evaluation I am skeptical that anybody _using_ reaction models would really want the duplicate reactions to be present, no? If you disagree another option would be to do the de-duplication inside the search algorithms.
- The return type being a `PredictionList` creates the need for extra wrapping/unwrapping. Do you think it could be possible to replace it with a plain list?

# Reviewing notes

The main changes in the PR are just 1-2 (which each get their own commit). The remainder of the commit is fixing some tests and downstream errors.

There is some refactoring I didn't do: for example, the `syntheseus_wrapper` stuff is made obsolete by this PR, but I left it in. If we decide to merge this I will of course fix this up and we can re-review the PR properly.

# Next steps?

If we like these changes, I can fix it up and do a proper PR. However, since it would almost certainly introduce breaking changes we would probably need to re-version it.


kmaziarz:
Thanks @AustinT, this looks very promising! Below are some comments on the various issues you discuss.

> Interfaces in `interface` allow duplicate reactants, but `search` assumes reactants are unique, leading to `set(r.reactants)` in a few places.

Yes, I recall we decided to allow duplicate reactants to be able to handle stoichiometry, although I'm not sure how common it is for the models to return non-unique reactants (equivalently, how common that is in training data). My hunch is that we should indeed keep the option to have non-unique reactants as you did in your changes (@mrwnmsr may have thoughts on this too). Your proposal to have a `unique_reactants` property sounds good; it's intuitive and would reduce code duplication.

> Two identical reactions with different scores or probabilities are now considered equal.

Sounds fine, I don't think we're ever comparing the prediction objects in single-step code. The deduplication logic in `eval_single_step` only looks at the reactants themselves, so it implicitly behaves in the way you describe.

> Move score/probability/etc to metadata, as they are not used by all models, and the purpose of having a metadata dict was specifically to avoid every reaction object carrying around a bunch of unused attributes filled with `None`.

SGTM. I'm also not sure if we need all of these bits of metadata, some are probably never set, but we can look into that in a separate PR.

> The `Prediction` class should be called `Reaction`, as this would be clearer and more specific.

Agreed. The `Prediction` class was originally made for single-step evaluation, and it was meant to abstract away whether we're predicting the reaction in the forward or backward direction; this is where the `input` / `output` naming comes from. In a broader context `Reaction` sounds more intuitive. There is a slight annoyance related to the typing of the `input` / `output` fields, as we assume in the backward direction the product (input) is a single `Molecule`, whereas in the forward direction there may be several products, so there the type is `Bag[Molecule]`. Not sure how to best resolve this, but my hunch is that the forward/backward abstractions for models/predictions are too complex given the utility they actually provide; maybe we don't need all this complexity and can just have a `Reaction` class which carries a `bool` denoting whether it was predicted in the forward or backward direction.

> To handle `num_results`, the cache keys are now `(mol, num_results)`, which may be redundant if the results for `num_results = n` are a prefix of the results for `num_results = m` for `n < m`.

It would often not be a prefix, as many models use some sort of beam search or reorder the predictions post-hoc. However, the output for `num_results = m` truncated to `n` results is probably at least as good of a prediction as the output for `num_results = n` in most cases, so there would be a bit of redundancy because we'd be simultaneously storing both a "lower quality" and a "higher quality" prediction. But, I don't think that is a problem: in the majority of cases (both in single-step evaluation and in search), `num_results` would be set to the same value throughout the "lifetime" of a model; I see adding `num_results` to the cache key as a way to make sure the behaviour is correct in an odd/unusual case, so I wouldn't worry about being a bit suboptimal.

>  It is unclear whether caching will mess up the evaluation script in some way. One solution would just be to turn off caching in this script.

I think caching should be turned off in single-step evaluation, because it would interfere with timing, and also due to higher memory usage. In fact, I wonder whether caching should really be on by default: if a user loads a model and computes predictions on their dataset, silently caching the outputs may make it seem as if there is a memory leak, as memory would keep increasing while all the data is stored in the model forever. On the other hand, we obviously want caching during search.

Maybe a compromise could be:
- caching is off by default, single-step evaluation keeps that
- `cli/search.py` turns caching on by default (this is already the case)
- if we want to be extra-safe, we could add a warning in `MinimalSearchAlgorithm` if the provided reaction model has caching turned off (for users that call the search primitives directly)

> Reaction duplicate removal now part of the model.

SGTM. I agree keeping the duplicates is mostly useful for detailed analysis but not so much for actually using the model for making predictions or running search. When adding the tutorial in #54, I did realize how awkward it is that we need a separate utility to call the model without getting duplicates; making this part of the models and on-by-default sounds like a good change.

> Effect of duplication removal on the evaluation script.

The single-step eval script already has a `skip_repeats` flag that controls this. The only difference would be that we would no longer need to handle deduplication in `get_results`, and instead would pass `skip_repeats` to the model.

> Default value for `num_results` provided at construction time, set to `100` by default.

Makes sense; for the default, maybe `50` would already be enough (the evaluation script asks for `100` by default to account for deduplication and make sure the top-`50` is computed more accurately, but in practice `50` results should be usually enough).

> The existence of the `PredictionList` class creates the need for extra wrapping/unwrapping.

Yes, I think having this class might be a bit overkill. The idea was to allow for attaching metadata that is not specific to a single predicted reaction and rather to the entire list of predictions, e.g. data about the model that made the prediction. If I recall correctly @mrwnmsr was a fan of this idea, but I think in practice we never actually set this metadata, and it does create more "fluff" in how the outputs are extracted (this can be seen in the tutorial notebook too; a lot of attribute accesses are needed just to get the outputs out). I would be happy with this being changed to just a plain list. If in the future we want to attach model-level metadata, we could also do it on top through a separate dataclass, rather than placing it in an object that is always created whenever a model is called.

> Bumping the package version.

These changes would indeed be breaking, and if we were already in the `1.*` version then the common practice would be to bump to `2.0.0`. However, we're in `0.*`, which (as far as I know) assumes breaking changes can happen freely, and so the next version could still be `0.*` e.g. `0.4.0`. We can release `1.0.0` once things stabilize a bit and we're more confident in the API.

AustinT:
Thanks for looking this over @kmaziarz! Overall I think the conclusion from this is that we should go ahead with a merger of some kind. What about the following changes (marking things you already seemed to agree to with ✅ ):

- **Reaction class**:
  - ✅ Rename base class `Prediction -> Reaction`.
  - Have `reactants/products` fields instead of input/output. Whether it was predicted in forwards/backward direction can be put into metadata if desired.
  - A subclass `SingleProductReaction` could be used in most of search and reaction prediction code.
  - ✅ Also provides `unique_reactants` alias
  - ✅ All probabilities/scores/etc go into `metadata` dict
- **PredictionList**: ✅ we eliminate this class and instead return lists of reactions. @mrwnmsr do you foresee any problems? (we could always subclass `list` to create a kind of `PredictionList` if needed)
- **Unified reaction model**:
  - ✅ Call signature contains `num_results` with sensible default.
  - ✅ Keep de-duplication and caching functionality in the model.
  - ✅ de-duplication enabled by default
  - ✅ cache keys are `(mol, num_results)` because redundancy shouldn't matter much in practice
  - ✅ caching off by default with a warning in search: this is a reasonable and intuitive compromise. A more complicated option could be `None` by default, which is set to `True/False` on the first call depending on whether it is called in a search algorithm. At this time I think that is overkill though and a warning would suffice.
- **Once merged**, bump to next minor version (`0.4.0`)

If you are happy with this, I think it would be cleanest to execute it in 2 separate PRs: one for the reaction class and one for the reaction model class. I can do this next week.

kmaziarz:
> Have `reactants` / `products` fields instead of `input` / `output`. 

SGTM, we just need to make sure this is properly taken into account in the eval script.

> A subclass `SingleProductReaction` could be used in most of search and reaction prediction code.

How would this subclass work? Would it change the type of `products`, or would it have a separate `product` property that asserts there is just one product and then returns it?

> Execute it in 2 separate PRs: one for the reaction class and one for the reaction model class.

Sounds good, the two PRs will already be relatively complex, so better not merge them 🙂 Potentially there could even be three sets of changes (an additional one for `PredictionList` -> `list`).

AustinT:
> > A subclass `SingleProductReaction` could be used in most of search and reaction prediction code.
> 
> How would this subclass work? Would it change the type of `products`, or would it have a separate `product` property that asserts there is just one product and then returns it?

There doesn't seem to be a clear best option to me. Maybe just changing the type of `product` is the easiest.

kmaziarz:
> Maybe just changing the type of `product` is the easiest.

I'm not sure how mypy will react to that, and it doesn't seem to be a good inheritance practice to change the type (as any instance of the subclass should always be a valid instance of the superclass).

AustinT:
Now redundant with the submission of #73.

PR #60, opened 2024-01-13
Update installation guide in docs

AustinT:
Clarify that version from GitHub might be unstable or have breaking changes.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,6 +3,12 @@ We support two installation modes:
 - *core installation* allows you to build and benchmark your own models or search algorithms
 - *full installation* also allows you to perform end-to-end search using the supported models
 
+There are also two installation sources:
+
+- *pip*, which provides the most recent released version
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -21,18 +27,20 @@ We support two installation modes:
...
     pip install -e .
     ```
 
-=== "Full (source)"
+=== "Full (GitHub)"
 
     ```bash
+    # NOTE: clone the repository first
     conda env create -f environment_full.yml
     conda activate syntheseus-full
```
kmaziarz, line 45: Actually, all of the instructions assume one has cloned the repository, as the `pip`-based ones still make use of the conda environments. This is why cloning the repo is mentioned in text below. So, I think we don't need these extra notes inside the instructions.

kmaziarz (approved):

docs/installation.md
AustinT: Fair point, I just removed it.

PR #62, opened 2024-01-14
Improve retro* implementation

AustinT:
Improved retro* implementation in several ways:

1. **Use better priority queue implementation**: the existing implementation did not track the set of nodes in the queue, and therefore could add a node to the queue multiple times. This generally caused the queue to be much larger than necessary, creating considerable slowdown in long searches. I fixed this by using a simple PQ implementation which allows for re-insertion.
2. **Track minimum cost route found so far**: this is essentially the "reaction number", but disregarding the search heuristic function. Tracking this could be useful for learning from self-play, and can be used to check whether the algorithm has found the minimum-cost route when the heuristic is admissible.
3. **Improve logging**: the queue size, graph size, and other details are logged in each iteration. I found this useful when running retro* for a large number of iterations.
4. **Fix edge case bug** where `reaction_number` depends on the `reaction_number_estimate` for all leaf nodes, including those which cannot be expanded (e.g. because they are too deep). This could cause the algorithm to pursue synthesis routes whose estimated cost can never be realized because one (or more) leaf nodes can never be expanded. I fixed this by only filling the `reaction_number_estimate` of nodes which can be expanded, and only including the `reaction_number_estimate` in the `reaction_number` calculation if it is present.

### DIFF
--- a/syntheseus/search/algorithms/best_first/base.py
+++ b/syntheseus/search/algorithms/best_first/base.py
@@ -68,13 +68,20 @@ class PriorityQueue:
                 # (if `None` is added to the queue there may be multiple `None` items)
                 del self._entry_finder[entry.item]
                 return entry
-        raise KeyError("pop from an empty priority queue")
+        raise IndexError("pop from an empty priority queue")
 
-    def __len__(self):
+    def __len__(self) -> int:
         """Length is number of unique items in the queue."""
         return len(self._entry_finder)
 
-    def raw_len(self):
+    def __contains__(self, item) -> bool:
+        return item in self._entry_finder
+
+    def get_priority(self, item) -> Any:
+        """Returns the priority of an item in the queue, raising KeyError if not found."""
+        return self._entry_finder[item].priority
+
+    def raw_len(self) -> int:
         return len(self._queue)
 
 
@@ -165,11 +172,8 @@ class GeneralBestFirstSearch(SearchAlgorithm[GraphType, int], Generic[GraphType]
             for updated_node in dict.fromkeys(new_nodes + list(nodes_updated)):
                 if self.node_eligible_for_queue(updated_node, graph):
                     updated_node_priority = self.priority_function(updated_node, graph)
-                    already_in_queue_at_correct_priority = (
-                        updated_node in queue._entry_finder
-                        and math.isclose(
-                            queue._entry_finder[updated_node].priority, updated_node_priority
-                        )
+                    already_in_queue_at_correct_priority = updated_node in queue and math.isclose(
+                        queue.get_priority(updated_node), updated_node_priority
                     )
 
                     if already_in_queue_at_correct_priority:
--- a/syntheseus/search/algorithms/best_first/base.py
+++ b/syntheseus/search/algorithms/best_first/base.py
@@ -136,9 +136,8 @@ class GeneralBestFirstSearch(SearchAlgorithm[GraphType, int], Generic[GraphType]
                 break
 
             # Take nodes from the priority queue until a node eligible for expansion is found
-            found_node = False
             num_popped = 0
-            while len(queue) > 0 and not found_node:
+            while len(queue) > 0:
                 pq_item = queue.pop_item()
                 num_popped += 1
 
@@ -149,9 +148,9 @@ class GeneralBestFirstSearch(SearchAlgorithm[GraphType, int], Generic[GraphType]
                 ), "priority is not up-to-date. This should not happen."
 
                 if self.node_eligible_for_queue(pq_item.item, graph):
-                    found_node = True
                     node = pq_item.item
-            if not found_node:
+                    break
+            else:
                 logger.log(log_level, "No eligible node found. Stopping search.")
                 break
 
--- a/syntheseus/search/algorithms/best_first/base.py
+++ b/syntheseus/search/algorithms/best_first/base.py
@@ -172,22 +172,33 @@ class GeneralBestFirstSearch(SearchAlgorithm[GraphType, int], Generic[GraphType]
                             queue._entry_finder[updated_node].priority, updated_node_priority
                         )
                     )
-                    if (
-                        not already_in_queue_at_correct_priority
-                        and updated_node_priority < math.inf
-                    ):
+
+                    if already_in_queue_at_correct_priority:
+                        # In this case re-inserting the node is redundant so we do nothing
+                        pass
+                    elif updated_node_priority < math.inf:
+                        # Main case: insert (or re-insert) node into queue
                         queue.push_item(updated_node, updated_node_priority)
+                    else:
+                        # Edge case: reached if new priority is inf.
+                        # In this case, remove the node from the queue if it was there.
+                        queue.remove_item(updated_node)
 
             # Log
             if logger_active:
+                logging_str = (
+                    f"Step: {step}, "
+                    f"Nodes affected during visit: {len(new_nodes)}, "
+                    f"Nodes updated: {len(nodes_updated)}, "
+                    f"Graph size: {len(graph)}, "
+                    f"Queue size: {len(queue)} (raw length: {queue.raw_len()}), "
+                    f"Num popped: {num_popped}, "
+                    f"Reaction model calls: {self.reaction_model.num_calls()}, "
+                    f"Node visited: {node}"
+                )
                 logger.log(
                     log_level,
-                    f"Step {step}:\tvisited node: {node}, "
-                    f"Nodes effected during visit: {len(new_nodes)}, "
-                    f"Nodes updated: {len(nodes_updated)}, "
-                    f"Graph size: {len(graph)}, ",
-                    f"Queue size: {len(queue)} (raw length: {queue.raw_len()}), ",
-                    f"Num popped: {num_popped}",
+                    logging_str,
                 )
 
         return step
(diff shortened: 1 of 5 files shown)


syntheseus/search/algorithms/best_first/base.py
```diff
@@ -2,17 +2,82 @@
...
+    they are just marked as invalid. This is why the queue length is not
+    necessarily the same as the number of unique items in the queue.
+    It also assumes that items are unique and hashable.
+    """
+
+    def __init__(self):
+        self._queue = []
+        self._entry_finder = {}
+        self._counter = itertools.count()  # to break ties in priority
+        self._REMOVED = None
```
kmaziarz, line 44: Is `self._REMOVED` always just an alias for `None`?

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -54,72 +119,86 @@ def _run_from_graph_after_setup(self, graph: GraphType) -> int:
...
+                assert math.isclose(
+                    pq_item.priority, self.priority_function(pq_item.item, graph)
+                ), "priority is not up-to-date. This should not happen."
+
+                if self.node_eligible_for_queue(pq_item.item, graph):
+                    found_node = True
+                    node = pq_item.item
+            if not found_node:
+                logger.log(log_level, "No eligible node found. Stopping search.")
+                break
```
kmaziarz, line 156: Could we equivalently use `while ... else` instead of having the `found_node` variable?
```suggestion
            num_popped = 0
            while len(queue) > 0:
                pq_item = queue.pop_item()
                num_popped += 1

                # Do a few checks
                assert pq_item.priority < math.inf, "inf priority should not be in the queue"
                assert math.isclose(
                    pq_item.priority, self.priority_function(pq_item.item, graph)
                ), "priority is not up-to-date. This should not happen."

                if self.node_eligible_for_queue(pq_item.item, graph):
                    node = pq_item.item
                    break
            else:
                logger.log(log_level, "No eligible node found. Stopping search.")
                break
```

AustinT: Yes, that's a good suggestion!

syntheseus/search/algorithms/best_first/base.py
AustinT: At the moment yes. Would you recommend having the option to override or customize this?

kmaziarz: What is the rationale for customizing this? Is it that currently it's not possible to put in an item that is equal to `None` because the `if entry.item is not self._REMOVED` logic will get confused, and setting `_REMOVED` to something else would allow that?

AustinT: Yes, that would be the rationale to customize it. But at the moment I didn't intend to put `None` into the priority queue so I thought it was fine to hard code it.

I was just suggesting the replacement because I didn't know why you made the comment.

kmaziarz: If we don't expect to put `None` items then my hunch would be to just directly use `None` for simplicity (and skip adding an "intermediate" in the form of `self._REMOVED`). Or we could also replace `if entry.item is not self._REMOVED` with `if entry.item in self._entry_finder` if we don't mind an additional dictionary lookup (which would remove the need to ever compare `entry.item` to the special removed value and we could use `None` even if some items are `None` to being with).

AustinT: This is a fair suggestion. I changed `self._REMOVED` to a hard-coded `None`. However, I don't think adding `None` to the queue will work properly with your suggested code change. Imagine in pseudo-code:

```
queue = PQ([])
queue.push(A, 0)  # queue is [(0, A)]
queue.change_priority(A, 1)  # queue will now be [(0, None), (1, A)]
queue.push(None, 2)  # queue will now be [(0, None), (1, A), (2, None)]
queue.pop()  # will return `(0, None)` which is incorrect
```

Instead I changed `if entry.item is not self._REMOVED` to `self._entry_finder.get(entry.item) == entry`

Does this seem like a good change?

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -2,16 +2,83 @@
...
+    priority: float
+    tie_breaker: int
+    item: Any = field(compare=False)
+
+
+class PriorityQueue:
+    """
+
+    Simple priority queue implementation which supports removing elements
+    and changing their priority.
```
kmaziarz, line 28: nit: extra endline
```suggestion
    """
    Simple priority queue implementation which supports removing elements
    and changing their priority.
```

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -2,16 +2,83 @@
...
+    queue. It also assumes that items are unique and hashable.
+    """
+
+    def __init__(self):
+        self._queue = []
+        self._entry_finder = {}
+        self._counter = itertools.count()  # to break ties in priority
+
+    def remove_item(self, item):
+        """Removes and item if it is present."""
```
kmaziarz, line 45:
```suggestion
        """Removes an item if it is present."""
```

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -2,16 +2,83 @@
...
+        """Removes an item with the lowest priority and returns it."""
+        while self._queue:
+            entry = heapq.heappop(self._queue)
+            if self._entry_finder.get(entry.item) == entry:
+                # ^^ ensures 1) item in entry finder and
+                # 2) item in entry finder has the same priority as the popped item
+                # (if `None` is added to the queue there may be multiple `None` items)
+                del self._entry_finder[entry.item]
+                return entry
+        raise KeyError("pop from an empty priority queue")
```
kmaziarz, line 72: nit: [`heapq` uses `IndexError` in this case](https://docs.python.org/3/library/heapq.html#heapq.heappop), which may be a slightly better fit

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -53,72 +120,85 @@ def _run_from_graph_after_setup(self, graph: GraphType) -> int:
...
+            num_popped = 0
+            while len(queue) > 0:
+                pq_item = queue.pop_item()
+                num_popped += 1
+
+                # Do a few checks
+                assert pq_item.priority < math.inf, "inf priority should not be in the queue"
+                assert math.isclose(
+                    pq_item.priority, self.priority_function(pq_item.item, graph)
+                ), "priority is not up-to-date. This should not happen."
```
kmaziarz, line 149:
```suggestion
                assert pq_item.priority < math.inf, "Inf priority should not be in the queue"
                assert math.isclose(
                    pq_item.priority, self.priority_function(pq_item.item, graph)
                ), "Priority in the queue should always be up-to-date"
```

syntheseus/search/algorithms/best_first/base.py
```diff
@@ -53,72 +120,85 @@ def _run_from_graph_after_setup(self, graph: GraphType) -> int:
...
+            # dict.fromkeys is to preserve order and uniqueness
+            for updated_node in dict.fromkeys(new_nodes + list(nodes_updated)):
+                if self.node_eligible_for_queue(updated_node, graph):
+                    updated_node_priority = self.priority_function(updated_node, graph)
+                    already_in_queue_at_correct_priority = (
+                        updated_node in queue._entry_finder
+                        and math.isclose(
+                            queue._entry_finder[updated_node].priority, updated_node_priority
+                        )
+                    )
```
kmaziarz, line 174: Maybe `PriorityQueue` could have `__contains__` and `get_priority` helpers implemented, so that the outside code doesn't need to access the private `_entry_finder` attribute?

syntheseus/search/algorithms/best_first/retro_star.py
```diff
@@ -123,12 +123,29 @@ def _run_retro_star_updates(
...
+            node.data.setdefault("retro_star_min_cost", math.inf)
+            node.data.setdefault("retro_star_reaction_number", math.inf)
             node.data.setdefault("retro_star_value", math.inf)
         nodes_to_update = set(cast(Collection[ANDOR_NODE], nodes))
 
         # NOTE: the following updates assume that depth is set correctly.
 
+        # Perform bottom-up update of `retro_star_min_cost`,
+        # sorting by decreasing depth and not updating children for efficiency
+        # (min costs depends only on children)
```
kmaziarz, line 135:
```suggestion
        # (min cost depends only on children)
```

kmaziarz (approved):
LGTM (apart from a few nitpicks).

A more general question: do you think the change (1) (better priority queue) will also speed up not-so-long searches, such as those shown in the original syntheseus paper (up to 10 minutes of search time), allowing them to perform slightly more steps? Or does this change really only come into play with searches much longer than that?

AustinT:
> LGTM (apart from a few nitpicks).
> 
> A more general question: do you think the change (1) (better priority queue) will also speed up not-so-long searches, such as those shown in the original syntheseus paper (up to 10 minutes of search time), allowing them to perform slightly more steps? Or does this change really only come into play with searches much longer than that?

You make good suggestions and I will implement your suggested code changes soon.

To answer your more general question: I think this will only affect very long searches. I made this change because in very long searches, the queue reached a length of 10^8 for a graph of size 10^6 which caused memory issues. For short searches it should not matter.

syntheseus/search/algorithms/best_first/base.py
AustinT: Good suggestions: done!

PR #63, opened 2024-01-15
Add new compromise reaction class

AustinT:
This PR adds a new compromise Reaction class in a new file: `syntheseus/interface/reaction.py`. The class has the features we discussed in #58 [^1]. The class also includes a basic test.

The new classes is not used anywhere at the moment[^2]. I will refactor the rest of the code to use it in another PR. I thought it would be test to propose the class in isolation first to avoid doing all the refactoring, then re-doing it if we decide to make additional changes.

[^1]: Except for the `unique_reactants` alias. After thinking about it a bit more, if users know to use `unique_reactants` then they could just as easily write `set(reactants)`.

[^2]: Except that I moved the `REACTION_SEPARATOR` constant to `reaction.py` and changed resulting imports. It seemed odd to have it in `molecule.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 ### Added
 
 - Add a new tutorial employing non-toy single-step models ([#54](https://github.com/microsoft/syntheseus/pull/54)) ([@kmaziarz])
+- Add new unified `Reaction` base class ([#63](https://github.com/microsoft/syntheseus/pull/63)) ([@austint])
 
 ### Fixed
 
--- a/syntheseus/interface/molecule.py
+++ b/syntheseus/interface/molecule.py
@@ -10,7 +10,6 @@ from rdkit import Chem
 from syntheseus.interface.typed_dict import TypedDict
 
 SMILES_SEPARATOR = "."
-REACTION_SEPARATOR = ">"
 
 
 class MoleculeMetaData(TypedDict, total=False):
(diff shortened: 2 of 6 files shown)


kmaziarz (approved):
> If users know to use `unique_reactants` then they could just as easily write `set(reactants)`.

This may be true, but I think another advantage of having `unique_reactants` is that it makes it more clear the `reactants` are _not_ unique. Otherwise I could imagine someone knowing they need unique reactants, but not realizing `reactants` are not already unique... Apart from that LGTM.

AustinT:
> > If users know to use `unique_reactants` then they could just as easily write `set(reactants)`.
> 
> This may be true, but I think another advantage of having `unique_reactants` is that it makes it more clear the `reactants` are _not_ unique. Otherwise I could imagine someone knowing they need unique reactants, but not realizing `reactants` are not already unique... Apart from that LGTM.

Good point. I added them to the base class. Do you think that is the appropriate place to add them @kmaziarz? There is some redundancy between the implementation in `MultiProductReaction` and `SingleProductReaction` but it is very small.

mrwnmsr (approved):

kmaziarz:
> I added them to the base class. Do you think that is the appropriate place to add them @kmaziarz? There is some redundancy between the implementation in `MultiProductReaction` and `SingleProductReaction` but it is very small.

Yes, what you've done seems reasonable.

AustinT:
> > I added them to the base class. Do you think that is the appropriate place to add them @kmaziarz? There is some redundancy between the implementation in `MultiProductReaction` and `SingleProductReaction` but it is very small.
> 
> Yes, what you've done seems reasonable.

Sounds good, I'll merge this then.

PR #65, opened 2024-01-24
Integrate the Graph2Edits model

kmaziarz:
This PR integrates Graph2Edits - a recently proposed model class which shows promising results on USPTO-50K. It includes a full model wrapper with a default checkpoint, as well as edits to docs and tutorials to mention the new model type.

Testing the integration uncovered a potential problem with some of the model classes: as the underlying repositories are often not set up for installability and use relative imports, creating LocalRetro loads a Python module called `models`; when one then tries to create Graph2Edits which also contains a module with that name, Python looks up the module in `sys.modules` rather than re-importing it from the new location. This leads to errors if the two models are used together, which was caught by our models test. To fix this, I refactored LocalRetro to place all modules under a top-level directory named `local_retro`, and released the updated code as a new version of `syntheseus-local-retro` on PyPI.

Finally, as import-related errors could potentially depend on the _order_ in which the two conflicting models are used, I also changed the models test to test each model type _twice_, covering _each pair of models in both orders_. Note that this does not affect running time, as most of the time is spent downloading model weights, and so re-testing a model again requires negligible extra time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,13 +35,13 @@ dev = [
   "pytest-cov",
   "pre-commit"
 ]
-chemformer = ["syntheseus-chemformer"]
-graph2edits = ["syntheseus-graph2edits"]
+chemformer = ["syntheseus-chemformer==0.1.1"]
+graph2edits = ["syntheseus-graph2edits==0.1.0"]
 local-retro = ["syntheseus-local-retro==0.4.0"]
-megan = ["syntheseus-megan"]
-mhn-react = ["syntheseus-mhnreact"]
-retro-knn = ["syntheseus-local-retro", "torch-scatter"]
-root-aligned = ["syntheseus-root-aligned"]
+megan = ["syntheseus-megan==0.1.0"]
+mhn-react = ["syntheseus-mhnreact==1.0.0"]
+retro-knn = ["syntheseus[local-retro]", "torch-scatter"]
+root-aligned = ["syntheseus-root-aligned==0.1.0"]
 all-single-step = [
   "syntheseus[chemformer,graph2edits,local-retro,megan,mhn-react,retro-knn,root-aligned]"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dev = [
 ]
 chemformer = ["syntheseus-chemformer"]
 graph2edits = ["syntheseus-graph2edits"]
-local-retro = ["syntheseus-local-retro"]
+local-retro = ["syntheseus-local-retro==0.4.0"]
 megan = ["syntheseus-megan"]
 mhn-react = ["syntheseus-mhnreact"]
 retro-knn = ["syntheseus-local-retro", "torch-scatter"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,13 +36,14 @@ dev = [
   "pre-commit"
 ]
 chemformer = ["syntheseus-chemformer"]
+graph2edits = ["syntheseus-graph2edits"]
 local-retro = ["syntheseus-local-retro"]
 megan = ["syntheseus-megan"]
 mhn-react = ["syntheseus-mhnreact"]
 retro-knn = ["syntheseus-local-retro", "torch-scatter"]
 root-aligned = ["syntheseus-root-aligned"]
 all-single-step = [
-  "syntheseus[chemformer,local-retro,megan,mhn-react,retro-knn,root-aligned]"
+  "syntheseus[chemformer,graph2edits,local-retro,megan,mhn-react,retro-knn,root-aligned]"
 ]
 all = [
   "syntheseus[viz,dev,all-single-step]"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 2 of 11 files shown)


AustinT (approved):
I think the `sys.module` modification is a bit hacky and that we should consider alternatives (more on this below). Modifying imports could have unintended side-effects for our users. However, apart from that the PR looks good so I am approving it. Up to you whether to merge @kmaziarz

What about these alternatives instead of your modification of `sys.modules`?

1. Rename poor model choices from `LocalRetro` + co: e.g. `models -> LocalRetroModels`. Since you control these repos this seems doable?
2. Remove relative imports in the code and all dependent repos.
3. Use `importlib.import_module` to resolve ambiguities (you can specify the exact import location here).

You suggested that 1-2 would "would make it hard to pull new changes from these repositories later on". I didn't totally understand why: can you elaborate?

kmaziarz:
> You suggested that 1-2 would "would make it hard to pull new changes from these repositories later on". I didn't totally understand why: can you elaborate?

Maybe not very hard, but currently the model repo forks mostly just either add or edit `setup.py` / `pyproject.toml` and leave the other files unchanged. Applying (1)/(2) means renaming/moving files, and changing most imports, which could lead to merge conflicts when we try to pull changes from upstream. This may not be a big concern though, especially in the case of older models like LocalRetro where we're not very likely to be pulling in upstream changes.

I'm not sure if (3) would work; it seems to be useful mostly for (a) cases when the module name is dynamic and (b) cases when we want to import without adding the module into the namespace. Here e.g. the internals of LocalRetro do things like `from models import ...`, so I think we do need `models` to be normally available for that to work...

AustinT:
> > You suggested that 1-2 would "would make it hard to pull new changes from these repositories later on". I didn't totally understand why: can you elaborate?
> 
> Maybe not very hard, but currently the model repo forks mostly just either add or edit `setup.py` / `pyproject.toml` and leave the other files unchanged. Applying (1)/(2) means renaming/moving files, and changing most imports, which could lead to merge conflicts when we try to pull changes from upstream. This may not be a big concern though, especially in the case of older models like LocalRetro where we're not very likely to be pulling in upstream changes.

That makes sense, but I agree that we probably aren't going to be pulling in changes from upstream. Personally if I wrote code that imported a generic name like `model` or `src` and somebody changed this in a fork I think that is totally sensible.

If I were making this PR I would probably just change the imports in our single-step model packages. Personally I find it really frustrating when libraries do hacky things under the hood and it causes bugs. These are usually really annoying to debug: an example that comes to mind is a bug Megan dealt with where `rdkit` monkey-patched some default HTML writer at runtime to render mols in pandas dataframes (which affected lots of other libraries). Of course, for most users what you've done will be fine, but I could imagine for example somebody experimenting locally and using a directory like `models` for their dev code, which would then be modified when they import LocalRetro to compare. However, I am not against merging it (we can always fix it later and it will most likely not impact anybody since relatively few people are using syntheseus right now). Up to you.

kmaziarz:
> I find it really frustrating when libraries do hacky things under the hood which has side effects.

Yeah, I guess you're right; hacking `sys.path` and the like worked out fine for the short-term if one just wants to use a single model in a relatively controlled environment, but (as we can see here) can cause issues long-term. I'll fix LocalRetro to make all of its imports be of the form `from local_retro...`, which will avoid the ambiguous module issue and also allow us to get rid of the `sys.path` manipulation in the model wrapper. This will be enough to get rid of the conflict with Graph2Edits; in the future we can perform this cleanup for the other models too and eventually get rid of all the import-related hackery 🙂

pyproject.toml
```diff
@@ -35,14 +35,15 @@ dev = [
...
-mhn-react = ["syntheseus-mhnreact"]
-retro-knn = ["syntheseus-local-retro", "torch-scatter"]
-root-aligned = ["syntheseus-root-aligned"]
+chemformer = ["syntheseus-chemformer==0.1.1"]
+graph2edits = ["syntheseus-graph2edits==0.1.0"]
+local-retro = ["syntheseus-local-retro==0.2.0"]
+megan = ["syntheseus-megan==0.1.0"]
+mhn-react = ["syntheseus-mhnreact==1.0.0"]
+retro-knn = ["syntheseus[local-retro]", "torch-scatter"]
+root-aligned = ["syntheseus-root-aligned==0.1.0"]
```
kmaziarz, line 44: Exact pins are needed because the single-step model repos are controlled by us and thus they are effectively a part of `syntheseus` (so there is no point in having the model versions be _not_ tied to `syntheseus` versions). If we did not pin, it would make future changes to single-step model repos disruptive to previously released `syntheseus` versions, e.g. if we release `syntheseus-megan 0.2.0` which includes breaking changes, all previous versions of `syntheseus` would automatically start pulling in the newer version of `syntheseus-megan`, whereas our intention would be for that new version to only affect the respectively updated version of `syntheseus`.

Because `syntheseus 0.3.0` did not pin, the just released version of `syntheseus-local-retro` does break LocalRetro in `syntheseus 0.3.0`. We could potentially add backwards compatibility hacks to prevent this, but I think it's better to just release `syntheseus 0.4.0` soon

kmaziarz: Actually, I think there is no breakage of `syntheseus 0.3.0`. I thought there was, but I made a mistake when releasing `syntheseus-local-retro`, which I now fixed. Nevertheless, the pins above are still useful, as in the future we may not be so lucky and an update to a single-step model repo may end up being breaking.

AustinT: That's a good idea. Looks good Kris!

AustinT (approved):
New changes look good Kris!

PR #66, opened 2024-01-26
Fix small issues in the integration of Graph2Edits

kmaziarz:
The recent integration of Graph2Edits in #65 missed two caveats:
- The `syntheseus-graph2edits` package did not specify that it depends on `joblib`, which was missed in initial testing as it's a rather common dependency that is pulled in by e.g. LocalRetro, so this issue only arises when Graph2Edits is used in isolation. I pushed version `0.2.0` that specifies the dependency and in this PR I bump the version requirement in syntheseus.
- In very rare cases (e.g. for the molecule with SMILES "OBr"), the Graph2Edits code fails with an unclear `IndexError`. This was not seen during evaluation on USPTO-50K, but happens occasionally during search. As these cases are very rare and there is no obvious fix, I resort to simply catching the exception and returning no proposed reactions in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


AustinT (approved):
LGTM!

PR #70, opened 2024-02-08
Rename value function to search heuristic

AustinT:
Retro*, MCTS, and proof number search all use some kind of function to estimate the "potential" of leaf nodes in the search graph and choose which node to expand at each iteration. In syntheseus we called this the "value function". However, in hindsight I think this is a poor name because:

1. Value function has a very specific meaning in reinforcement learning: it is the expected reward obtained from following a policy and *none* of the uses in syntheseus fit that. At best they are *estimates* of the value function.
2. From the RL setting, value function has the connotation of "higher is better", but for algorithms like retro* lower values are better. This is probably confusing to users.

To improve clarity, this PR renames value function to "search heuristic" whenever referring to a "general" value function rather than something which is specifically a value function (e.g. in MCTS). I think the term "search heuristic" is both more accurate and less confusing than value function, and at least to me the term implies that it will be interpreted differently for different algorithms.

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -56,8 +56,8 @@ logger = logging.getLogger(__file__)
 class RetroStarConfig:
     max_expansion_depth: int = 10
 
-    value_function_class: str = "ConstantNodeEvaluator"
-    value_function_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.0})
+    search_heuristic_class: str = "ConstantNodeEvaluator"
+    search_heuristic_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.0})
 
     and_node_cost_fn_class: str = "ReactionModelLogProbCost"
     and_node_cost_fn_kwargs: Dict[str, Any] = field(default_factory=dict)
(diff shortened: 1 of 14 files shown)


syntheseus/cli/search.py
```diff
@@ -67,8 +67,8 @@ class RetroStarConfig:
 class MCTSConfig:
     max_expansion_depth: int = 20
 
-    value_function_class: str = "ConstantNodeEvaluator"
-    value_function_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.5})
+    search_heuristic_class: str = "ConstantNodeEvaluator"
+    search_heuristic_kwargs: Dict[str, Any] = field(default_factory=lambda: {"constant": 0.5})
```
kmaziarz, line 71: In the context of MCTS, isn't "value function" a widely accepted term? Also, doesn't the use of this term here roughly match what it means in RL? I worry that switching from a usual term to a generic one may actually confuse someone who has used MCTS in the past.

mrwnmsr: I agree. Let's leave "value function" as it is. I also don't think it is confusing.
We can add a comment or documentation to clarify that for some algorithms the value function is a (heuristic) negative cost function

mrwnmsr (changes requested):
Let's leave "value function" as it is. I also don't think it is confusing.
We can add a comment or documentation to clarify that for some algorithms the value function is a (heuristic) negative cost function

AustinT:
> Let's leave "value function" as it is. I also don't think it is confusing. We can add a comment or documentation to clarify that for some algorithms the value function is a (heuristic) negative cost function

Do you mean it is not a confusing term _in general_ or just for the specific MCTS example which Kris raised above?

If you think "value function" is better than "search heuristic" in general then let's not change it; it is just a name. In that case I can close this PR.

syntheseus/cli/search.py
kmaziarz: Yeah, I mean we could also rename "value function" to something else in Retro*-style algorithms, and keep it in MCTS-style algorithms, if that helps? I don't have a strong preference either way though.

AustinT:
Ok let's not merge this. I will close the PR.

PR #71, opened 2024-02-15
Add numpy 2 compatibility check to ruff

AustinT:
Numpy 2.0 will be released soon and it is advised to check all packages which depend on numpy to ensure that they won't break when this happens (as there are a few API changes). Thankfully there is a ruff check for this. I added it to our `pyproject.toml` and bumped the ruff version in our pre-commit check.


kmaziarz (approved):
Makes sense, thanks for looking into this!

The new version of `ruff` seems to not like our `pyproject.toml` configuration:
```
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
```

AustinT:
> The new version of `ruff` seems to not like our `pyproject.toml` configuration:

Good catch, I fixed this. This error did not show up for me when ran pre-commit though. Where did you see it? Some log file?

kmaziarz:
> Where did you see it? Some log file?

At the bottom of the pre-commit step log [here](https://github.com/microsoft/syntheseus/actions/runs/7915870195/job/21608593354). But I've also seen this warning locally when running freshly-installed `ruff` from the command line.

PR #73, opened 2024-02-15
Unify backward reaction models from search and reaction_prediction

AustinT:
Following discussion from #58, this PR unifies the base reaction model class from `search` and `reaction_prediction`, finally allowing all code to use a single reaction model base class in all places.

The new class lives in `syntheseus.interface.models.py` and is mainly an extended version of the earlier `ReactionModel` class to include:

- caching (disabled by default)
- de-duplication of reactions (enabled by default)
- call signature where `num_results` is optional and defaults to 100
- tracking number of calls
- a reset function to clear the cache / call count

The following modifications were made to use the new class:

- all pointers in the search code to `search.reaction_models.BackwardReactionModel` were redirected to `interface.models.BackwardReactionModel`
- Tests of caching / de-duplication which previously lived under `tests/search/*` were moved to `tests/interface/test_models.py`
- all models in `reaction_prediction` were modified to override `_get_reactions` instead of `__call__` (otherwise it would avoid interacting with the cache)
- delete the syntheseus wrapper class (no longer necessary). Search CLI now just directly uses the single-step model created from config.
- Add a warning to search algorithms if reaction model caching is switched off. Also ensured caching is turned on for all models used during the tests in `tests/search/*`

After this, `search/reaction_models` still contained code for toy reaction models. I thought this was a bit out of place so I moved it to `reaction_prediction/inference/toy_models.py` and included the word "toy" in the class names to emphasize that these are toy models. I also moved the tests. This allowed `search/reaction_prediction` to be deleted entirely.


syntheseus/cli/search.py
```diff
@@ -165,14 +163,7 @@ def run_from_config(config: SearchConfig) -> Path:
...
     # Load the single-step model
-    base_model = get_model(config, batch_size=1, num_gpus=int(config.use_gpu))  # type: ignore
-
-    # Set up the search algorithm
-    search_rxn_model = SyntheseusBackwardReactionModel(
-        cast(BackwardReactionModel, base_model),
-        num_results=config.num_top_results,
-        use_cache=config.reaction_model_use_cache,
-    )
+    search_rxn_model = get_model(config, batch_size=1, num_gpus=int(config.use_gpu))  # type: ignore
```
kmaziarz, line 166: Don't we still have to set `num_results` and `use_cache` here?

CHANGELOG.md
```diff
@@ -11,13 +11,13 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 - Suppress the remaining noisy logs and warnings coming from single-step models ([#53](https://github.com/microsoft/syntheseus/pull/53)) ([@kmaziarz])
 - Improve efficiency and logging of retro* algorithm ([#62](https://github.com/microsoft/syntheseus/pull/62)) ([@austint])
-- Remove `PredictionList` objects, replacing them with `Sequence[Prediction]` ([#61](https://github.com/microsoft/syntheseus/pull/61)) ([@austint])
+- Merged reaction and reaction model base classes in "search" and "reaction_predction" ([#63](https://github.com/microsoft/syntheseus/pull/63), [#67](https://github.com/microsoft/syntheseus/pull/67), [#73](https://github.com/microsoft/syntheseus/pull/73)) ([@austint])
+- Reaction models now return `Sequence[Reaction]` instead of `PredictionList` objects ([#61](https://github.com/microsoft/syntheseus/pull/61)) ([@austint])
```
kmaziarz, line 15: Change to imperative mood to match the changelog convention
```suggestion
- Merge reaction and reaction model base classes in `search` and `reaction_predction` ([#63](https://github.com/microsoft/syntheseus/pull/63), [#67](https://github.com/microsoft/syntheseus/pull/67), [#73](https://github.com/microsoft/syntheseus/pull/73)) ([@austint])
- Make reaction models return `Sequence[Reaction]` instead of `PredictionList` objects ([#61](https://github.com/microsoft/syntheseus/pull/61)) ([@austint])
```

syntheseus/cli/search.py
AustinT: That was unclear to me: since these are already in the config, are they not handled automatically by `get_model`? Maybe I am misunderstanding the internals of this method though.

kmaziarz: Yes, but `get_model` sees the `config` as a `BackwardModelConfig`, so hyperparameters added on top of that in `SearchConfig` are not accessed. Thus, `num_top_results` and `reaction_model_use_cache` are currently unused, and the defaults set there are ignored; running `search.py` out-of-the-box does produce your warnings about caching being turned off. For something to be passed to the single-step model it has to be captured in the `model_kwargs` argument to `BackwardModelConfig`, so e.g. caching can be turned on from CLI via:
```
model_kwargs="{'use_cache': True}"
```

I'm wondering what the best approach here is. We could in theory take `num_top_results` and `reaction_model_use_cache` and put those into `model_kwargs`, but that would lead to having two paths for specifying the same hyperparameter. Instead, we could delete those two arguments, assuming number of results and caching will be configured through `model_kwargs`, but we need to make sure the defaults are set right (currently they are not). Overriding defaults is tricky to marry with `SearchConfig` inheriting from `BackwardModelConfig`, but we could switch to having a field instead; something like
```python
class SearchConfig:
    model_config: BackwardModelConfig = BackwardModelConfig(
        model_kwargs={"use_cache": True, "default_num_results": 50}
    )
    ...
```

This would resolve the confusion and set the correct defaults. The only remaining thing would be that changing `default_num_results` sounds like a relatively common thing to explore, and setting it via CLI would be a bit clunky (also might lead to accidentally overriding the default for `use_cache`). If we want to improve this, we could modify `ModelConfig` to add `use_cache` and `default_num_results` explicitly (the `model_kwargs` bit was mostly intended for kwargs that are unique to a particular model class).

What do you think?

AustinT: Ah, I see the additional complication. None of these solutions seem ideal. What about just doing:

```python
search_rxn_model = get_model(config, batch_size=1, num_gpus=int(config.use_gpu))  # type: ignore
search_rxn_model._use_cache = config.reaction_model_use_cache
search_rxn_model.default_num_results = config.num_top_results
search_rxn_model.reset()
```

The only "bad" thing we are doing here is setting the private `_use_cache` attribute. However, since the model has not been called at this stage and we reset it immediately nothing unexpected should happen.

kmaziarz: Looks a bit hacky but also seems okay. A variant of this would be to add a `**kwargs` argument to `get_model` (and pass that directly to the model `__init__`), so that we can avoid changing things after construction:
```python
search_rxn_model = get_model(
    config,
    batch_size=1,
    num_gpus=int(config.use_gpu)),
    use_cache=config.reaction_model_use_cache,
    default_num_results=config.num_top_results,
)
```

AustinT: Good idea. I just pushed a commit with the following changes:

1. As you suggest, add **model_kwargs to get_model (with a check to ensure duplicate kwargs are not supplied)
2. Create the model with `get_model` as you suggested in your most recent comment, supplying these kwargs
3. Just in case we want it for the future, I added an option to change `use_cache` when calling `model.reset()`. This means that the "hacky" solution I suggested above could be done in a non-hacky way in the future.

What do you think of these changes?

kmaziarz: The newest changes look good to me!

kmaziarz:
> On the topic of `get_results` in `eval_single_step.py`, it now repeats the logic for deduplicating the outputs; I think instead of having `skip_repeats` be an argument to `get_results` we should just pass the respective config option when creating the model via `get_model`.

Digging this up, I think it's actually more than duplicated logic: the `skip_repeats` argument in `eval_single_step.py` will turn the deduplication logic in the script on/off, but that will have no effect due to deduplication already being turned on in the model itself. I think we can again make use of `get_model` accepting arbitrary kwargs and do
```python
model = get_model_fn(config, remove_duplicates=config.skip_repeats)
```
After that we can get rid of the deduplication logic from `eval_single_step.py` entirely.

AustinT:
> After that we can get rid of the deduplication logic from eval_single_step.py entirely.

Good point, I will make this change shortly (just going for lunch now).

AustinT:
> After that we can get rid of the deduplication logic from eval_single_step.py entirely.

I pushed another commit which does this. What do you think?

kmaziarz (approved):

PR #74, opened 2024-02-27
Simplify class hierarchy for reaction objects

AustinT:
In #63 the new reaction class used the attribute "product" (no 's') to refer to both the single product in `SingleProductReaction` and the set of products in `MultiProductReaction`s. This is a bit confusing, so this PR refactors the class structure so that `MultiProductReaction` is simply called `Reaction`, and `SingleProductReaction` is a subclass which a `product` property. The resulting class structure contains no generics and is much simpler.

### DIFF
--- a/syntheseus/interface/reaction.py
+++ b/syntheseus/interface/reaction.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from abc import abstractmethod
 from dataclasses import dataclass, field
-from typing import Collection, Generic, Optional, TypeVar
+from typing import Collection, Optional, TypeVar
 
 from syntheseus.interface.bag import Bag
 from syntheseus.interface.molecule import SMILES_SEPARATOR, Molecule
@@ -28,37 +27,6 @@ class ReactionMetaData(TypedDict, total=False):
     ground_truth_match: bool  # whether this reaction matches ground truth
 
 
-@dataclass(frozen=True, order=False)
-class Reaction(Generic[ReactantsType]):
-    """General reaction class."""
-
-    # The molecule that the prediction is for and the predicted output:
-    reactants: ReactantsType = field(hash=True, compare=True)
-    identifier: Optional[str] = field(default=None, hash=True, compare=True)
-
-    # Dictionary to hold additional metadata.
-    metadata: ReactionMetaData = field(
-        default_factory=lambda: ReactionMetaData(),
-        hash=False,
-        compare=False,
-    )
-
-    @property
-    @abstractmethod
-    def reaction_smiles(self) -> str:
-        pass
-
-    @property
-    @abstractmethod
-    def unique_reactants(self) -> set[Molecule]:
-        pass
-
-    @property
-    @abstractmethod
-    def unique_products(self) -> set[Molecule]:
-        pass
-
-
 def combine_mols_to_string(mols: Collection[Molecule]) -> str:
     """Produces a consistent string representation of a collection of molecules."""
     return SMILES_SEPARATOR.join([mol.smiles for mol in sorted(mols)])
@@ -70,23 +38,17 @@ def reaction_string(reactants_str: str, product_str: str) -> str:
 
 
 @dataclass(frozen=True, order=False)
-class _MultiProductBase:
-    """Dummy class to avoid non-default argument following default argument"""
-
+class Reaction:
+    reactants: Bag[Molecule] = field(hash=True, compare=True)
     products: Bag[Molecule] = field(hash=True, compare=True)
+    identifier: Optional[str] = field(default=None, hash=True, compare=True)
 
-
-@dataclass(frozen=True, order=False)
-class MultiProductReaction(
-    Reaction[Bag[Molecule]],
-    _MultiProductBase,
-):
-    @property
-    def reaction_smiles(self) -> str:
-        return reaction_string(
-            reactants_str=combine_mols_to_string(self.reactants),
-            product_str=combine_mols_to_string(self.products),
-        )
+    # Dictionary to hold additional metadata.
+    metadata: ReactionMetaData = field(
+        default_factory=lambda: ReactionMetaData(),
+        hash=False,
+        compare=False,
+    )
 
     @property
     def unique_reactants(self) -> set[Molecule]:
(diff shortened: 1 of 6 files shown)


AustinT:
Reviewing note: in addition to running tests, I also ran the single-step eval CLI and it worked for me, so hopefully this PR shouldn't break anything?

syntheseus/interface/reaction.py
```diff
@@ -30,12 +29,11 @@ class ReactionMetaData(TypedDict, total=False):
...
 
 @dataclass(frozen=True, order=False)
-class Reaction(Generic[ReactantsType, ProductType]):
+class Reaction(Generic[ReactantsType]):
     """General reaction class."""
 
     # The molecule that the prediction is for and the predicted output:
     reactants: ReactantsType = field(hash=True, compare=True)
-    product: ProductType = field(hash=True, compare=True)
     identifier: Optional[str] = field(default=None, hash=True, compare=True)
```
kmaziarz, line 37: I think our reaction types hierarchy is a bit overcomplicated given what it tries to accomplish. What about having the `products` attribute in the base class, getting rid of `MultiProductReaction` entirely, and having `SingleProductReaction` add only a `@property` that asserts there is exactly one product and returns it (without adding any actual attributes)?

kmaziarz: Additionally, perhaps `Reaction` doesn't need to be generic at all; we may just want to fix the types of `reactants` and `products` to `Bag[Molecule]` (and the `product` property would return `Molecule`).

AustinT: I thought about that, but the disadvantage of this approach would be that construction is more awkward: you need would write `SingleProductReaction(products={mol}, ...)` (i.e. always wrapping mol in a set or Bag) and the restriction that there is only one mol would only be checked when you try to *access* it, which is not ideal in my opinion.

I do agree that it could be simplified though, for example by making the class no longer generic. What about just removing the generic part?

AustinT: For example, what about the following class structure @kmaziarz ?

```python
class _BaseReaction:
    reactants: Bag[Molecule]
    ... # identifier / metadata

class Reaction(_BaseReaction):
    products: Bag[Molecule]

class SingleProductReaction(_BaseReaction)
    product: Molecule
```

(note: `@dataclass` flags and string methods are not shown)

This is less complicated than my current solution, but still allows `SingleProductReaction` objects to be created without an empty `Bag` wrapper. We could even add a `products` property to `SingleProductReaction` if we wanted to.

kmaziarz: > what about the following class structure

This looks fine, it is a step in the right direction.

> I thought about that, but the disadvantage of this approach would be that construction is more awkward: you would write `SingleProductReaction(products={mol}, ...)` (i.e. always wrapping mol in a set or bag) and the restriction that there is only one mol would only be checked when you try to _access_ it, which is not ideal in my opinion.

Hmm, yes; the later could be tackled with a check in `__post_init__`, but the former may be tricky to get around.

AustinT: > This looks fine, it is a step in the right direction.

What would you propose instead? I agree it's not ideal but none of the options seem ideal...

kmaziarz: Yeah, I agree, none of what we discussed seems ideal. Here's another approach: what about overriding the auto-generated `__init__` in `SingleProductReaction`:

```python
@dataclass
class Reaction:
    reactants: Bag[Molecule]
    products: Bag[Molecule]


@dataclass
class SingleProductReaction(Reaction):
    def __init__(self, reactants: Bag[Molecule], product: Molecule) -> None:
        super().__init__(reactants=reactants, products=Bag([product]))

    @property
    def product(self) -> Molecule:
        assert len(self.products) == 1  # Guaranteed in `__init__`.
        return self.products[0]
```
Wrapping the product into a `Bag` would only appear in one place (`__init__` in `SingleProductReaction`), and that also guarantees that `len(self.products) == 1` for any instance of `SingleProductReaction` (so there would be no way to make the `assert` fail without some low-level Python hackery).

AustinT: That approach makes sense, but is a bit complex. It almost just seems like a complicated way to ensure that `SingleProductReaction`s have a `products` attribute (with an s). Is that even a desirable outcome?

If it is just about the class hierarchy then we can just simply it using a `typing.Protocol`. Would that be better?

kmaziarz: > It almost just seems like a complicated way to ensure that `SingleProductReaction`s have a products attribute (with an s). Is that even a desirable outcome?

My thinking was that this would allow us to process both kinds of reactions in a unified way. Currently `eval_single_step.py` if-cases this; it would be nice if we could compare the produced reaction to the one from the dataset by just comparing `reactants` and `products` directly.

AustinT:
Is the main goal to be able to compare SingleProductReactions with MultiProductReactions directly without changing the type? If so, then I could see the case for a wrapper. Otherwise if the benefit is just avoiding an if statement do you think it justifies the additional complexity?
________________________________

kmaziarz:
> Is the main goal to be able to compare `SingleProductReaction`s with `MultiProductReaction`s directly without changing the type? If so, then I could see the case for a wrapper.

I think `eval_single_step.py` would anyway have to compare the reactants/products directly, because one side of the comparison will be the richer reaction sample coming from the dataset, and the other side would be a reaction coming from the model, so directly they will never equal anyway.

I wouldn't call the `SingleProductReaction` in my last proposal a "wrapper"; it is just a way to encode in the type system that we're dealing with an instance of `Reaction` _of a particular kind_ (where the products list has length exactly 1). In that sense a `SingleProductReaction` is just a special instance of a `Reaction`, and any code that just handles general reactions could treat `SingleStepReaction`s as `Reaction`s. The different `__init__` and the `product` property were simply a convenience thing.

To me, this seems simpler than the alternative approaches, and also avoids handling inheritance in the presence of default fields in a special way (because the `SingleStepReaction` subclass is not adding any new fields). That being said, we can also go with [your last proposal](https://github.com/microsoft/syntheseus/pull/74#discussion_r1505751318). If we add a `products` property to `SingleStepReaction` then functionally the outcome will be very similar (biggest difference being a more complex inheritance hierarchy with an "incomplete" base class).

AustinT:
I think my main issue is that it seems wasteful to wrap a single molecule in a Bag and then repeatedly unwrap it just for the sake of a nicer class structure.

What is your main hesitation about the complex class hierarchy: is it that it is not easy to read the code and understand which classes have which attributes, or is it the type hierarchy itself? If we were didn't need the dummy classes would you think it is cleaner?
________________________________

kmaziarz:
> I think my main issue is that it seems wasteful to wrap a single molecule in a Bag and then repeatedly unwrap it just for the sake of a nicer class structure.

Ah, yeah, that part is a bit redundant.

> What is your main hesitation about the complex class hierarchy: is it that it is not easy to read the code and understand which classes have which attributes, or is it the type hierarchy itself? If we were didn't need the dummy classes would you think it is cleaner?

A bit of all of these, I suppose: the code is harder to read, there is no sensible base class that could be used to process both kinds of reaction types uniformly, and one needs the dummy base classes to set up the inheritance in the presence of default fields. But in general your approach is also OK, and getting rid of generics will make things simpler anyway, so we can also just go with that instead of debating further.

syntheseus/interface/reaction.py
```diff
@@ -1,15 +1,13 @@
...
-from abc import abstractmethod
 from dataclasses import dataclass, field
-from typing import Collection, Generic, Optional, TypeVar
+from typing import Collection, Optional, TypeVar
 
 from syntheseus.interface.bag import Bag
 from syntheseus.interface.molecule import SMILES_SEPARATOR, Molecule
 from syntheseus.interface.typed_dict import TypedDict
 
 ReactantsType = TypeVar("ReactantsType")
```
kmaziarz, line 10: Is this now unused?

kmaziarz (approved):
LGTM!

(Maybe also update the PR title/description to match the final contents of the PR, as those will end up in the merge commit)

syntheseus/interface/reaction.py
AustinT: Good catch. Just pushed a fix to remove it.

AustinT:
> (Maybe also update the PR title/description to match the final contents of the PR, as those will end up in the merge commit)

Good suggestion: just changed this.

PR #76, opened 2024-03-04
Further reduce code duplication between reaction classes

kmaziarz:
Our data loading code produces `ReactionSample` objects, which (as of #74) are very similar to `Reaction`s, except for holding a few extra things. These two classes also duplicate some of the functionality, e.g. utils for stringifying a `Bag` of `Molecule`s. This PR makes an initial attempt at deduplicating these two classes, making `ReactionSample` inherit from `Reaction`.

Longer term we could potentially combine the two classes entirely (we can defer this to a future PR to speed up the release of `v0.4.0`).

### DIFF
--- a/syntheseus/interface/molecule.py
+++ b/syntheseus/interface/molecule.py
@@ -85,4 +85,9 @@ class Molecule:
 
 
 def molecule_bag_to_smiles(mols: Bag[Molecule]) -> str:
+    """Combine SMILES strings of molecules in a `Bag` into a single string.
+
+    For two bags that represent the same multiset of molecules this function will return the same
+    result, because iteration order over a `Bag` is deterministic (sorted using default comparator).
+    """
     return SMILES_SEPARATOR.join(mol.smiles for mol in mols)
--- a/syntheseus/interface/molecule.py
+++ b/syntheseus/interface/molecule.py
@@ -7,6 +7,7 @@ from typing import Optional, Union
 
 from rdkit import Chem
 
+from syntheseus.interface.bag import Bag
 from syntheseus.interface.typed_dict import TypedDict
 
 SMILES_SEPARATOR = "."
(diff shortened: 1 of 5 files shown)


syntheseus/interface/reaction.py
```diff
@@ -56,11 +51,18 @@ def unique_reactants(self) -> set[Molecule]:
     def unique_products(self) -> set[Molecule]:
         return set(self.products)
 
+    @property
+    def reactants_combined(self) -> str:
+        return molecule_bag_to_smiles(self.reactants)
+
+    @property
+    def products_combined(self) -> str:
+        return molecule_bag_to_smiles(self.products)
```
kmaziarz, line 60: This was the original name of these properties in `ReactionSample`, but maybe we should call them differently to make it clear that they convert the reactants/products to a string? One could imagine "combining" a `Bag[Molecule]` into a single (disconnected) `Molecule`, which is _not_ what is done here...

syntheseus/interface/molecule.py
```diff
@@ -81,3 +82,7 @@ def rdkit_mol(self) -> Chem.Mol:
         if "rdkit_mol" not in self.metadata:
             self.metadata["rdkit_mol"] = Chem.MolFromSmiles(self.smiles)
         return self.metadata["rdkit_mol"]
+
+
+def molecule_bag_to_smiles(mols: Bag[Molecule]) -> str:
+    return SMILES_SEPARATOR.join(mol.smiles for mol in mols)
```
AustinT, line 88: I noticed you are not sorting the molecules (while my previous string method sorted them). Is this because the contents of a `Bag` will always be sorted? If so, maybe clarify in a docstring that this should produce a consistently-ordered string.

syntheseus/interface/reaction.py
AustinT: Yeah I think the names `reactants_str` and `products_str` would be more fitting.

AustinT (approved):
In general LGTM, but it would be nice to rename `reactants_combined` and `products_combined` to something more clear.

PR #78, opened 2024-03-05
Update docs and tutorials

kmaziarz:
This PR combines a few updates to our docs and tutorials:
- updating the "Quick Start" tutorial after recent refactoring
- explaining in the installation section that the version of the docs used should match how `syntheseus` was installed
- adding basic documentation for the `search` CLI endpoint

Docs that include the changes from this PR are already up under [microsoft.github.io/syntheseus/dev/](https://microsoft.github.io/syntheseus/dev/).

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -48,7 +48,7 @@ There are also two installation sources:
 !!! note
 
     Make sure you are viewing the version of the docs matching your `syntheseus` installation.
-    Select the `x.y.z` version you installed (go [here](https://microsoft.github.io/syntheseus/stable/) for the latest one) if you used `pip`,
+    Select the `x.y.z` version you installed if you used `pip` (go [here](https://microsoft.github.io/syntheseus/stable/) for the latest one),
     or [dev](https://microsoft.github.io/syntheseus/dev/) if you installed `syntheseus` directly from GitHub.
 
 Core installation includes only minimal dependencies (no ML libraries), while full installation includes all supported models and also dependencies for visualization/development.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -47,9 +47,9 @@ There are also two installation sources:
 
 !!! note
 
-    Make sure you are viewing the version of the docs that matches your `syntheseus` installation.
-    Select [stable](https://microsoft.github.io/syntheseus/stable/) (0.3.0) if using the latest version available via `pip`,
-    and [dev](https://microsoft.github.io/syntheseus/dev/) if installing directly from GitHub.
+    Make sure you are viewing the version of the docs matching your `syntheseus` installation.
+    Select the `x.y.z` version you installed (go [here](https://microsoft.github.io/syntheseus/stable/) for the latest one) if you used `pip`,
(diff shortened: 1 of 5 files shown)


docs/installation.md
```diff
@@ -45,6 +45,12 @@ There are also two installation sources:
     pip install -e ".[all]"
     ```
 
+!!! note
+
+    Make sure you are viewing the version of the docs that matches your `syntheseus` installation.
+    Select [stable](https://microsoft.github.io/syntheseus/stable/) (0.3.0) if using the latest version available via `pip`,
```
kmaziarz, line 51: Mentioning `0.3.0` here as otherwise someone might be confused seeing how `stable` is not an option in the version selector. A downside is that this will have to be updated with each release (or we could figure out an option to automatically show what the latest version is here, but this may require plugging in a custom mkdocs extension).

AustinT: We could simply not explicitly state 0.3.0 and just write something like:

> Make sure you are viewing the version of the docs that matches your `syntheseus` installation.
> This will generally be a version tag (e.g. 0.X.Y). Select *stable* if you are using the latest version available from pip, or *dev* if installing directly from GitHub.

Would that be a reasonable solution?

AustinT (approved):
LGTM! I tried running search following the instructions in the docs and it worked for me.

docs/installation.md
kmaziarz: Except "select stable" doesn't make sense, because that is not present in the version selector? But I could rephrase to avoid saying the word "stable" entirely

AustinT: Ah true. That's weird, since the link `https://microsoft.github.io/syntheseus/stable/` works and points to the correct version. We could just say "select the correct version" or dev and leave it to the user then?

kmaziarz: > That's weird, since the link https://microsoft.github.io/syntheseus/stable/ works and points to the correct version.

Yeah, by default navigating to the docs will go to `stable` which will redirect to the version with that alias.

kmaziarz: I changed it; how about now?

docs/installation.md
```diff
@@ -47,9 +47,9 @@ There are also two installation sources:
 
 !!! note
 
-    Make sure you are viewing the version of the docs that matches your `syntheseus` installation.
-    Select [stable](https://microsoft.github.io/syntheseus/stable/) (0.3.0) if using the latest version available via `pip`,
-    and [dev](https://microsoft.github.io/syntheseus/dev/) if installing directly from GitHub.
+    Make sure you are viewing the version of the docs matching your `syntheseus` installation.
+    Select the `x.y.z` version you installed (go [here](https://microsoft.github.io/syntheseus/stable/) for the latest one) if you used `pip`,
```
AustinT, line 51:
```suggestion
    Select the `x.y.z` version you installed if you used `pip` (go [here](https://microsoft.github.io/syntheseus/stable/) for the latest one),
```

AustinT: (I think this word order makes the sentence slightly easier to parse)

AustinT (approved):

PR #79, opened 2024-03-06
Add tutorial for integrating custom models and improve README

kmaziarz:
This PR adds a new tutorial that shows how to integrate a custom reaction model and run search with it (for demonstration purposes it uses a toy model which is a simplified version of a model we use in tests). Additionally, it also slightly improves the README by adding links at the top to docs, CLI reference, tutorials and our paper. The first three technically all link to somewhere in the docs, but having them spelled out at the top may make it more clear to users that these things (especially tutorials) do exist. Finally, I also apply a small cleanup to `.gitignore` to remove an entry which is no longer relevant.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,3 @@ MANIFEST
 # Unit test / coverage reports
 .coverage
 .coverage.*
-
-# Cloned single-step model repositories
-syntheseus/reaction_prediction/environments/external/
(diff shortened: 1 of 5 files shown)


kmaziarz:
See [here](https://microsoft.github.io/syntheseus/dev/tutorials/custom_model/) for the new tutorial and [here](https://github.com/microsoft/syntheseus/tree/kmaziarz/add-custom-model-tutorial) for how README would look like after this PR.

AustinT (approved):
This new tutorial looks really good Kris. Thanks for adding it!

PR #81, opened 2024-03-08
Make basic classes from `interface` importable from top-level

AustinT:
Almost any script using syntheseus in a non-trivial way will need to import a handful of essential classes from `syntheseus.interface`. This PR makes them importable from top-level, allowing users to write

```python
from syntheseus import Molecule, Reaction, Bag
```

instead of

```python
from syntheseus.interface.molecule import Molecule
from syntheseus.interface.reaction import Reaction
from syntheseus.interface.bag import Bag
```

Honestly I often forget where to import these things from, and if I find this annoying, I imagine syntheseus's users would feel similarly.


AustinT:
## TODO before merging:

- Add CHANGELOG entry

kmaziarz (approved):
Looks good, I also noticed this when writing the tutorials. I guess there's no need to update internal code to use the simplified imports, but maybe we should update the tutorials here?

(Also, in the future we might want to expose some more classes like this, e.g. make it possible to import the model wrappers from `syntheseus.reaction_prediction` instead of `syntheseus.reaction_prediction.inference`, but we can take things one step at a time and think/decide on those later in future PRs.)

AustinT:
> Looks good, I also noticed this when writing the tutorials. I guess there's no need to update internal code to use the simplified imports, but maybe we should update the tutorials here?

Sounds good: I updated the tutorials as you suggested, and added a CHANGELOG entry.

PR #83, opened 2024-03-09
Add random search algorithm

AustinT:
This PR adds random search as a simple baseline for both MolSet and AND/OR graphs. Specifically, "random" here means at each iteration, a node is chosen uniformly at random (from the set of all un-expanded nodes) for expansion. This PR also adds basic tests (but does include a hand-worked example, because the random behaviour makes writing such a test futile). Also, some tests in the base class were skipped because random search seems to be very inefficient.


syntheseus/search/algorithms/random.py
```diff
@@ -0,0 +1,65 @@
...
+            self.set_node_values(set(new_nodes) | {node}, graph)
+            for n in new_nodes:
+                if self.can_expand_node(n, graph):
+                    expandable_nodes.add(n)
+
+            if logger_active:
+                logger.log(
+                    log_level,
+                    f"Step {step}: node {node} expanded, created {len(new_nodes)} new nodes. "
+                    f"Num expandable nodes: {len(expandable_nodes)}",
```
kmaziarz, line 50:
```suggestion
                    f"Num expandable nodes: {len(expandable_nodes)}.",
```

kmaziarz (approved):
LGTM.

I noticed the new algorithm is not integrated into the search CLI, but I'm not sure if we need that? If we did integrate it, then we would also get extra test coverage via the CLI test from #80, but then again I'm not sure if we care. Seems fine to merge this as it is I think.

AustinT:
> I noticed the new algorithm is not integrated into the search CLI

Let's not add it for now: this is not really an algorithm which we expect people to use in practice; it is mostly just a baseline for papers (where people will probably use it in a larger python script). If we want it then we can add it later?

PR #85, opened 2024-04-05
Add option to stop search after search graph exceeds a certain size

AustinT:
This option is called `limit_graph_nodes`. I also added a basic test for this.


syntheseus/tests/search/algorithms/test_base.py
```diff
@@ -157,6 +157,45 @@ def test_limit_reaction_model_calls(
...
+        self,
+        retrosynthesis_task6: RetrosynthesisTask,
+        limit: int,
+    ) -> None:
+        """
+        Test that limiting the number of nodes in the graph works as intended.
+        The algorithm should run until the node limit is reached, and then stop
+        (without adding *too many* extra nodes).
+
+        retrosynthesis_task6 is chosen because it can create a very large graph.
```
kmaziarz, line 171:
```suggestion
        `retrosynthesis_task6` is chosen because it can create a very large graph.
```

syntheseus/tests/search/algorithms/test_base.py
```diff
@@ -157,6 +157,45 @@ def test_limit_reaction_model_calls(
...
+        alg = self.setup_algorithm(
+            reaction_model=retrosynthesis_task6.reaction_model,
+            mol_inventory=retrosynthesis_task6.inventory,
+            limit_graph_nodes=limit,
+            limit_iterations=int(1e6),  # a very high limit, but avoids MCTS warnings
+        )
+        output_graph, _ = alg.run_from_mol(retrosynthesis_task6.target_mol)
+
+        # The algorithm will stop running when the graph size meets or exceeds the limit.
+        # HOWEVER, since multiple nodes are added during each expansion, the node count might
```
kmaziarz, line 184: No need to shout 😄
```suggestion
        # However, since multiple nodes are added during each expansion, the node count might
```

kmaziarz (approved):
LGTM!

syntheseus/search/algorithms/base.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         self,
         limit_iterations: int = INT_INF,
         limit_reaction_model_calls: int = INT_INF,
+        limit_graph_nodes: int = INT_INF,
```
kmaziarz, line 73: Do we also want to have this as an option in the search CLI?

AustinT: Good idea. I just pushed another commit to do this. Can you check whether you think I've modified the file in all the right places?

AustinT: (@kmaziarz , sorry forgot to tag you)

syntheseus/cli/search.py
```diff
@@ -187,6 +189,7 @@ def run_from_config(config: SearchConfig) -> Path:
                 "limit_reaction_model_calls",
                 "limit_iterations",
                 "prevent_repeat_mol_in_trees",
+                "limit_graph_nodes",
```
kmaziarz, line 192: Just keeping the same order as above 😄
```suggestion
                "limit_graph_nodes",
                "prevent_repeat_mol_in_trees",
```

(This "suggestion" feature of GitHub is making me really nitpicky, because with it I assume the cost of applying the change for the creator of the PR is effectively zero, leading me to propose changes I wouldn't bother with otherwise...)

kmaziarz (approved):
Thanks for updating the search CLI - I think what you've done should be enough to make this work.

PR #86, opened 2024-04-09
Improve typing of reaction metadata

kmaziarz:
Following a discussion in #80, this PR adjusts the type hints for reaction metadata to be `ReactionMetaData` instead of the generic `Dict[str, Any]` to allow for more precise type checking.

### DIFF
--- a/syntheseus/reaction_prediction/inference/root_aligned.py
+++ b/syntheseus/reaction_prediction/inference/root_aligned.py
@@ -123,10 +123,10 @@ class RootAlignedModel(ExternalBackwardReactionModel):
             assert new_score <= 0.0
 
             kwargs_list.append(
-                {
-                    "original_score": score,  # type: ignore[typeddict-unknown-key]
-                    "best_pos": best_pos,  # type: ignore[typeddict-unknown-key]
-                    "total_rr": total_rr,  # type: ignore[typeddict-unknown-key]
+                {  # type: ignore[typeddict-unknown-key]
+                    "original_score": score,
+                    "best_pos": best_pos,
+                    "total_rr": total_rr,
                     "score": new_score,
                 }
             )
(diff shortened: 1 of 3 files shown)


syntheseus/reaction_prediction/inference/root_aligned.py
```diff
@@ -121,14 +121,15 @@ def _build_kwargs_from_scores(self, scores: List[float]) -> List[Dict[str, Any]]
...
-            kwargs_list.append(metadata)
+
+            kwargs_list.append(
+                {
+                    "original_score": score,  # type: ignore[typeddict-unknown-key]
+                    "best_pos": best_pos,  # type: ignore[typeddict-unknown-key]
+                    "total_rr": total_rr,  # type: ignore[typeddict-unknown-key]
+                    "score": new_score,
+                }
+            )
```
kmaziarz, line 132: As expected, after improving the type hint `mypy` complains about the keys that `ReactionMetaData` does not know about. For now I just added appropriate `type: ignore` statements as it didn't seem useful to add RootAligned-specific keys to the underlying class, but I'm open to trying alternative approaches.

AustinT (approved):
LGTM! Not sure why a test is still failing though, since it looks like you fixed the underlying error.

kmaziarz:
> Not sure why a test is still failing though, since it looks like you fixed the underlying error.

I'm confused too, it seems as if the `type: ignore` thing doesn't work in Python 3.7? 😕

> Does this merit a CHANGELOG entry in your opinion?

I think we can probably leave this one out, as the change is quite minor (typing only) and mostly internal (not very important for users unless somebody integrates their own model _and_ does so using our utilities).

syntheseus/reaction_prediction/inference/root_aligned.py
AustinT: Maybe make a plain dict and use `typing.cast` to force mypy to treat it as a `ReactionMetaData` object? That should hopefully work around whatever is happening with our python 3.7 tests.

kmaziarz: Actually just moving `# type: ignore[typeddict-unknown-key]` to apply to the entire dictionary (and not a single entry) seems to work 🙂

PR #87, opened 2024-04-10
Add a workflow to release a new stable version

kmaziarz:
This PR adds a new GitHub Action that partially automates the process of releasing a new stable version of `syntheseus`. It tags the latest commit with an appropriate version tag and builds the docs (using a generalized version of the docs building pipeline).

I could not fully test the new action yet since it does not exist on `main`, but in case any issues arise when releasing v0.4.0 the action will be updated in a follow-up PR.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,6 +3,11 @@ name: Docs
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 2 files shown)


mrwnmsr (approved):
looks good to me

PR #92, opened 2024-04-26
Fix issues with GLN

kmaziarz:
The GLN model is the only model class that uses a bespoke environment based on Python 3.7 instead of the shared environment that uses Python 3.9. Due to this, it is not tested in CI, and is generally prone to bugs being introduced over time.

After trying to run the GLN model using the weights uploaded to figshare, I found that the upload was incomplete, and some files were missing from the checkpoint, preventing the model from being loaded. On top of this, there were a few small issues in the wrapper that accumulated over time, which also needed to be addressed to get the model to run again. This PR changes the weights link to a complete upload and also fixes the small issues in the model wrapper.

### DIFF
--- a/syntheseus/reaction_prediction/inference/gln.py
+++ b/syntheseus/reaction_prediction/inference/gln.py
@@ -57,6 +57,10 @@ class GLNModel(ExternalBackwardReactionModel):
 
             self.model = RetroGLN(self.model_dir, chkpt_path)
 
+    @property
+    def name(self) -> str:
+        return "GLN"
+
     def get_parameters(self):
         return self.model.gln.parameters()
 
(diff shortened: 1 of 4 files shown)


syntheseus/reaction_prediction/inference/gln.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, *args, dataset_name: str = "schneider50k", **kwargs) -> None:
 
             self.model = RetroGLN(self.model_dir, chkpt_path)
 
+    @property
+    def name(self) -> str:
+        return "GLN"
```
kmaziarz, line 62: The base class does `self.__class__.__name__.removesuffix("Model")`, but `removesuffix` does not exist in Python 3.7...

mrwnmsr (approved):

AustinT (approved):
LGTM

PR #93, opened 2024-05-03
Release v0.4.1

kmaziarz:
Since the release of `v0.4.0` we fixed major issues with two of the single-step model wrappers (GLN used to not run at all, while RetroKNN would run with degraded performance due to erroneous weights). This PR marks the release of `v0.4.1` to make sure these fixes are pushed to PyPI.


AustinT (approved):
LGTM Kris!

PR #94, opened 2024-06-07
Shift the pandas dependency to the model packages

kmaziarz:
The `pandas` dependency was so far part of our conda environment without any pin, but it seems more recent versions can cause some issues in LocalRetro when combined with `rdkit`, causing CI to fail. Given that `pandas` is available from PyPI, this PR removes the dependency from the environment and instead adds it (with a `<2.2` constraint) to the `syntheseus-local-retro` package.


AustinT (approved):
LGTM!

PR #96, opened 2024-08-13
Fix constructor arguments in `ParallelReactionModel`

kmaziarz:
One of the more rarely used capabilities of our codebase is running multi-GPU inference during single-step evaluation. Testing this part of code requires having a GPU, and so `ParallelReactionModel` was not covered by unit tests, and it broke at some point during refactoring of the model classes. This PR adds simple tests for it and fixes the issue.


AustinT (approved):
LGTM Kris!

kmaziarz:
By the way, with this PR the overall test coverage went from 91% to 92% 😄

PR #97, opened 2024-08-13
Make it easier to run search with custom models

kmaziarz:
Our single-step evaluation CLI defines the evaluation-specific arguments in `BaseEvalConfig`, and then creates `EvalConfig` by combining `BaseEvalConfig` and `BackwardModelConfig` as base classes. This design allows one to inject custom models by defining a modified `BackwardModelConfig`, then creating a modified `EvalConfig` (by mixing in `BaseEvalConfig`), and passing that to the syntheseus code (relying on a bit of duck typing). Internally, this allowed us to easily benchmark internal models using the shared eval script from syntheseus. This PR adapts the search CLI to follow a similar convention, so that one can also easily integrate custom models into search.

### DIFF
--- a/syntheseus/cli/main.py
+++ b/syntheseus/cli/main.py
@@ -1,10 +1,14 @@
 import sys
+from typing import Callable, Dict
 
 from syntheseus.cli import eval_single_step, search
 
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


AustinT (approved):
LGTM Kris (although I didn't test it myself). The modifications seem simple enough that I don't anticipate any problems arising from it.

PR #98, opened 2024-08-22
Reuse search results when given a partially filled directory

kmaziarz:
When running search on many targets using preemptible/unreliable compute, it may happen that the run fails having completed a fraction of targets. If such a run is restarted, it will generate a new timestamped output directory and start from scratch. This PR adds a flag `append_timestamp_to_dir`, which when set to `False` turns off timestamping, meaning that a restarted run would use the same directory as the old one. On top of this, it implements logic to skip targets which were completed earlier; a target that was partially completed (for example the search statistics were saved but it crashed when dumping the graph) will be purged and recomputed. Finally, the PR also includes a few small tweaks to the surrounding code: adding guards around visualization imports (to make sure search can be ran with the base environment without `graphviz`) and deleting an `__init__.py` file in an old empty directory.

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -29,7 +29,7 @@ import yaml
 from omegaconf import MISSING, DictConfig, OmegaConf
 from tqdm import tqdm
 
-from syntheseus.interface.molecule import Molecule
+from syntheseus import Molecule
 from syntheseus.reaction_prediction.inference.config import BackwardModelConfig
 from syntheseus.reaction_prediction.utils.config import get_config as cli_get_config
 from syntheseus.reaction_prediction.utils.misc import set_random_seed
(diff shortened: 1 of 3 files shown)


syntheseus/cli/search.py
```diff
@@ -46,7 +46,12 @@
 from syntheseus.search.mol_inventory import SmilesListInventory
 from syntheseus.search.node_evaluation import common as node_evaluation_common
 from syntheseus.search.utils.misc import lookup_by_name
-from syntheseus.search.visualization import visualize_andor, visualize_molset
+
+try:
+    # Try to import the visualization code, which will work only if `graphviz` is installed.
+    from syntheseus.search.visualization import visualize_andor, visualize_molset
+except ModuleNotFoundError:
```
mrwnmsr, line 53: is there additional reporting we want to do here, e.g. log/print that this error was excepted?

AustinT: One idea would be:

```python
try:
    # Try to import the visualization code, which will work only if `graphviz` is installed.
    from syntheseus.search.visualization import visualize_andor, visualize_molset
    VIZ_CODE_IMPORTED = True
except ModuleNotFoundError:
    VIZ_CODE_IMPORTED = False
```

Then, in the main CLI somewhere, maybe write:

```python
assert VIZ_CODE_IMPORTED or config.num_routes_to_plot == 0
```

AustinT (approved):
LGTM, except for my comment under Marwin's

PR #99, opened 2024-08-31
Fix `dictify` to handle torch tensors

mrwnmsr:
PR fixes that the `dictify` function does not handle tensors, which can be needed in some occasions.

### DIFF
--- a/syntheseus/reaction_prediction/utils/misc.py
+++ b/syntheseus/reaction_prediction/utils/misc.py
@@ -7,7 +7,7 @@ from dataclasses import fields, is_dataclass
 from itertools import islice
 from os import devnull
 from typing import Any, Dict, Iterable, Iterator, List, Optional
-
+from torch import Tensor
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


AustinT (approved):
LGTM. I think the ruff error in pre-commit might be fixed by introducing a blank line after line 9 (to separate built-in modules from third-party ones).

kmaziarz (changes requested):
I think this will fail the tests, because the core of the library is supposed to run without optional dependencies such as `torch`. But equally, I don't think we really need to support tensors here. This came up because one of our internal models had a bug and was returning its prediction probability as a tensor. The solution isn't to support tensors in `dictify` but rather fix the model 🙂 All of the models in syntheseus return their outputs as plain Python datatypes and I think we can keep it that way.

AustinT:
Good catch Kris, sorry I completely forgot about not supporting torch as a top-level import. Lesson: don't review PRs when I am very tired 🥲
________________________________

mrwnmsr:
lets close this then

PR #102, opened 2024-09-11
Extend single-step evaluation with stereo-agnostic results

kmaziarz:
This PR adds an `include_results_modulo_stereo` flag to the single-step evaluation script to additionally compute metrics (top-k accuracy, MRR) under a relaxed notion of success which ignores stereochemistry information. This can be useful to assess what fraction of errors made by the model are due to a mismatch in chirality. Additionally, this PR also slightly tidies up the printing of eval results.

### DIFF
--- a/syntheseus/cli/eval_single_step.py
+++ b/syntheseus/cli/eval_single_step.py
@@ -35,6 +35,7 @@ from syntheseus.interface.models import (
     ReactionType,
 )
 from syntheseus.interface.reaction import Reaction, SingleProductReaction
+from syntheseus.reaction_prediction.chem.utils import remove_stereo_information_from_reaction
 from syntheseus.reaction_prediction.data.dataset import (
     DataFold,
     DiskReactionDataset,
@@ -77,6 +78,7 @@ class BaseEvalConfig:
     skip_repeats: bool = True  # Whether repeated results should be skipped
     num_gpus: int = 1  # Number of GPUs to use (or 0 if running on CPU)
     num_dataset_truncation: Optional[int] = None  # Subset size to evaluate on
+    include_results_modulo_stereo: bool = False  # Whether to include stereo-agnostic results
 
     # Fields for saving and printing results
     print_idxs: List[int] = field(default_factory=lambda: [1, 3, 5, 10, 20, 50])
@@ -121,6 +123,8 @@ class EvalResults:
     mean_num_predictions: float
     median_num_predictions: float
     model_time_total: ModelTimingResults
+    match_modulo_stereo_top_k: Optional[List[float]] = None
+    match_modulo_stereo_mrr: Optional[float] = None
     back_translation_top_k: Optional[List[float]] = None
     back_translation_mrr: Optional[float] = None
     predictions: Optional[List[Sequence[Reaction]]] = None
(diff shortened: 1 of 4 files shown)


syntheseus/cli/eval_single_step.py
```diff
@@ -397,21 +423,27 @@ def compute_metrics_from_config(
...
-        if f.name not in ("model_info", "top_k", "predictions", "back_translation_predictions"):
-            print(f"{f.name}: {getattr(results, f.name)}")
+        if f.name not in ("model_info", "predictions", "back_translation_predictions"):
+            if f.name.endswith("top_k"):
+                top_k = getattr(results, f.name)
+                if top_k is not None:
+                    # Extract the top-k for the chosen values of `k`; other values are not printed.
+                    chosen_top_k_results[f.name] = {x: top_k[x - 1] for x in config.print_idxs}
+            else:
+                print(f"{f.name}: {getattr(results, f.name)}")
```
AustinT, line 437: Just to be clear, this part of the PR is unrelated to stereochemistry, right? It just seems to neaten up the printing.

syntheseus/reaction_prediction/chem/utils.py
```diff
@@ -29,6 +29,19 @@ def remove_atom_mapping(smiles: str) -> str:
...
+
+def remove_stereo_information_from_reaction(reaction: Reaction) -> Reaction:
+    return Reaction(
+        reactants=Bag([remove_stereo_information(mol) for mol in reaction.reactants]),
+        products=Bag([remove_stereo_information(mol) for mol in reaction.products]),
+        identifier=reaction.identifier,
+        metadata=reaction.metadata,
+    )
+
+
```
AustinT, line 44: Will this correctly handle the case when a `SingleProductReaction` is given? It looks like this code will downgrade it to a basic `Reaction` object. Could this cause the reaction comparison logic in `eval_single_step.py` to fail?

AustinT (approved):
LGTM (apart from a small concern about type conversion).

syntheseus/cli/eval_single_step.py
kmaziarz: Slightly related: this PR adds `match_modulo_stereo_top_k`, which (similarly to the pre-existing `back_translation_top_k`) would be printed verbosely on the CLI instead of limiting to just 1/3/5/10/20/50. I tidied up the printing in a general way which also fixes this oversight for `back_translation_top_k`.

syntheseus/reaction_prediction/chem/utils.py
kmaziarz: Yeah, it would downgrade to a `Reaction`. The reason why the eval script still worked is because `remove_stereo_information_from_reaction` was applied to _both_ sides of the comparison so both were getting downgraded. I now changed `remove_stereo_information_from_reaction` so that it retains the type of the input, and added tests to verify this works as expected.

More generally, I find it annoying that a `Reaction` that happens to have a single product is not considered equal to a corresponding `SingleProductReaction`. I think `SingleProductReaction` should be mostly syntax sugar for easier construction and type checking. But if we want to change this (i.e. make the reaction comparison ignore the class and only look at the internals), this would be a separate PR anyway.

AustinT: That's a good point: I agree that changing the comparison between `Reaction` and `SingleProductReaction` objects makes sense, but it would make the most sense as a separate PR. Would there be a natural way to do this that doesn't involve manually writing a comparison for each attribute?

kmaziarz: Not sure if there's a way to tell a dataclass to do this natively, but I'd just write a simple comparator in the base `Reaction` class that compares the fields; should be simple enough?

AustinT: Sounds good!

PR #104, opened 2024-10-15
Add an option to set time limit for route extraction

kmaziarz:
Our route extraction utilities can behave very differently depending on the structure of the underlying graph, and thus it's hard to predict how long extracting a given number of routes will take. In some cases, even the time between yielding one route and the next can be quite substantial. To allow for more control over route extraction, this PR adds a `max_time_s` argument to those utilities, which can be used to cap the total time spent. This time constraint is enforced in every iteration (as opposed to on every route found), which allows for following the limit much more closely.


AustinT (approved):
This is a good feature to have Kris- route extraction hanging for long periods of time is really frustrating. Thanks for adding this!

Issue #106, opened 2024-11-13
pytest failures (most likely my environment)


DocMinus:
When running the pytest I get a number of Fs. Most likely due to the install? Had to do some manual hands-on, the out-of-box pip install didn't work.
I am running on WSL2 Unbuntu 22 (with Nvidia GPU).
What I did differently after the first pip install due to errors after bailing out due to torch_scatter:

mamba install -c pyg -c conda-forge torch-scatter
pip install "syntheseus[all]"
pip install typing-extensions==4.6.0 (since 4.5.0 not combatible)

the first run of tests conked out due to graphviz?
Thus I manually installed that via
sudo apt install graphviz
and rerun
pip install "syntheseus[all]"
which reinstalled typing_extenstions 4.5.0.
Running pytests then led to this
```
syntheseus/tests/cli/test_cli.py ..F...F..F...F..F...F..F...F.                                                                                                                                                                                                                                                        [  5%]
syntheseus/tests/cli/test_eval_single_step.py .............................                                                                                                                                                                                                                                           [ 11%]
syntheseus/tests/cli/test_search.py .                                                                                                                                                                                                                                                                                 [ 11%]
syntheseus/tests/interface/test_bag.py .                                                                                                                                                                                                                                                                              [ 11%]
syntheseus/tests/interface/test_models.py ......                                                                                                                                                                                                                                                                      [ 12%]
syntheseus/tests/interface/test_molecule.py .........                                                                                                                                                                                                                                                                 [ 14%]
syntheseus/tests/interface/test_reaction.py .....                                                                                                                                                                                                                                                                     [ 15%]
syntheseus/tests/interface/test_toy_models.py ..                                                                                                                                                                                                                                                                      [ 15%]
syntheseus/tests/reaction_prediction/chem/test_utils.py ...                                                                                                                                                                                                                                                           [ 16%]
syntheseus/tests/reaction_prediction/data/test_dataset.py ..............                                                                                                                                                                                                                                              [ 19%]
syntheseus/tests/reaction_prediction/inference/test_models.py ..F.......F.....F.......F...                                                                                                                                                                                                                            [ 24%]
syntheseus/tests/reaction_prediction/utils/test_misc.py ....                                                                                                                                                                                                                                                          [ 25%]
syntheseus/tests/reaction_prediction/utils/test_parallel.py ..                                                                                                                                                                                                                                                        [ 25%]
[...]

PR #107, opened 2024-11-17
Correctly specify dependency on typing_extensions

AustinT:
Syntheseus previously depended on typing_extensions without specifying this in the `pyproject.toml` file, causing a fresh install to fail the tests. This PR fixes this by:

1. Wrapping the second import of typing_extensions in an if/else block to only run if python < 3.8 is used (otherwise an import from `typing` can be used directly). This was already present in the first import of typing_extensions.
2. Modify pyproject.toml to specify that `typing_extensions` is required, but _only if python < 3.8_.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
     "omegaconf",            # reaction_prediction
     "rdkit",                # reaction_prediction, search
     "tqdm",                 # reaction_prediction
+    "typing_extensions ; python_version < '3.8'",  # for TypedDict and Protocol imports
 ]
 
 [project.optional-dependencies]
--- a/syntheseus/interface/bag.py
+++ b/syntheseus/interface/bag.py
@@ -1,7 +1,13 @@
+import sys
 from collections.abc import Collection
 from typing import Generic, Iterable, Iterator, TypeVar
 
-from typing_extensions import Protocol
+# mypy-compatible import recommended here:
+# https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
 
 
 class Comparable(Protocol):


AustinT:
(I noticed this issue when investigating #106 )

AustinT:
@kmaziarz the tests are failing in this PR due to a segmentation fault, but this doesn't seem very related to the PR? I've tried re-running 3 times now but it still won't pass. Not sure what's going on here...

kmaziarz:
> Modify pyproject.toml to specify that `typing_extensions` is required, but only if python < 3.8

My thinking was that we should drop support for Python 3.7 soon anyway; we only maintained it due to being required by GLN, but as of #103 this is no longer the case.

Perhaps we should release `0.5.0` to have a final numbered version that still supports Python 3.7, and then bump the version requirement after that?

> Tests are failing in this PR due to a segmentation fault, but this doesn't seem very related to the PR

Yeah, doesn't seem related to the PR. It also happens from `main`, and seems to also be caused by `torch-scatter`... 😆 I'll need to take a closer look at what's going on

kmaziarz:
Update: I don't think the environment issues are _really_ caused by `torch_scatter`, and rather by the fact that installing `syntheseus[all]` now reinstalls the `torch` despite it already being present in the env. This is due to another package updating upstream. I'll send a fix soon although we may want to set things up a bit more robustly in the long term...

AustinT:
Great Kris, thanks for looking into this! I agree about not supporting python 3.7 anymore- that makes our lives a lot easier. Releasing a final python 3.7 version and then dropping support sounds good.

kmaziarz:
The issue should be fixed as of #108. I think this is not everything, as I also get problems with building `torch-scatter` as reported in the other issue. I'll look into that too, but I think that's separate, and doesn't seem to happen in CI. At the very least #108 should unblock your PR, which I imagine we still may want to merge in order to get a fully working `0.5.0` version (before we drop support for Python 3.7 and strip out some of the changes from your PR)?

AustinT:
Sounds good, once #108 is merged lets re-run CI, merge this if there are no issues, make a new release, then drop python 3.7 support!

kmaziarz:
> Sounds good, once #108 is merged lets re-run CI, merge this if there are no issues, make a new release, then drop python 3.7 support!

I've rebased the PR onto `main` now; lets see if the CI passes this time around.

AustinT:
Looks like it passes @kmaziarz. Should I merge?

kmaziarz (approved):

kmaziarz:
Seems good to merge, maybe just add a CHANGELOG entry.

AustinT:
Sorry, didn't see your comment re CHANGELOG

AustinT:
@kmaziarz fixed CHANGELOG in #109

PR #110, opened 2024-11-26
Install `torch_scatter` from conda instead of pip

kmaziarz:
As highlighted by DocMinus in #106, our installation instructions can fail due to issues with `torch_scatter`. This is because our current setup is essentially equivalent to `pip install torch-scatter`, which compiles the library from source, and so depends on having the right versions of `nvcc` and `gcc`. Alternatively, we could specify an exact wheel from `https://data.pyg.org/whl/`, but that makes things a bit less portable and harder to switch between different Python versions. Instead, this PR shifts the `torch_scatter` dependency to use `pyg`'s conda channel as opposed to installation through `pip`.

### DIFF
--- a/environment_full.yml
+++ b/environment_full.yml
@@ -3,13 +3,15 @@ channels:
   - defaults
   - conda-forge
   - pytorch
-  - dglteam       # LocalRetro, RetroKNN
+  - dglteam                 # LocalRetro, RetroKNN
+  - pyg                     # RetroKNN
 dependencies:
-  - dgl-cuda11.3  # LocalRetro, RetroKNN
-  - faiss-gpu     # RetroKNN
+  - dgl-cuda11.3            # LocalRetro, RetroKNN
+  - faiss-gpu               # RetroKNN
   - numpy
   - pip
   - python==3.9.7
   - pytorch=1.10.2=py3.9_cuda11.3_cudnn8.2.0_0
-  - rdchiral_cpp  # MHNreact
+  - pytorch-scatter==2.0.9  # RetroKNN
+  - rdchiral_cpp            # MHNreact
   - rdkit=2021.09.4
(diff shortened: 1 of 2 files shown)


AustinT (approved):
I agree with taking torch-scatter out of the pyproject dependencies, but this will effectively mean that `pip install "syntheseus[all]"` still won't install a completely working version. To minimize user frustration, what about adding a `try/except` statement [here](https://github.com/microsoft/syntheseus/blob/8f14b7d4920eb4d6bbbfb38c1e1a76fcf016ac58/syntheseus/reaction_prediction/models/retro_knn.py#L61) with a user message saying roughly:

> torch scatter is a pain to install so we didn't include it as a dependency, install it yourself via conda or something else

That seems like the only part that will fail without torch-scatter, but if I missed something maybe add it there too.

AustinT:
> > I agree with taking torch-scatter out of the pyproject dependencies, but this will effectively mean that pip install "syntheseus[all]" still won't install a completely working version.
> 
> Are you referring to a situation when somebody installs syntheseus but doesn't use our environment (the latter already containing `torch_scatter` from conda as of this PR)? In that case they would be missing several other dependencies as well; things that we get from conda are not listed in project requirements already (e.g. `torch`, `faiss-gpu`, `dgl`, ...).

Yes, that's what I was referring to. Very good point though, I completely forgot about not having torch/dgl/etc mentioned in pyproject at all 😅 

In that case, let's not bother adding a message. Happy for you to merge as is.

kmaziarz:
> I completely forgot about not having torch/dgl/etc mentioned in pyproject at all 😅

Yeah, I'm not sure what the best setup would be here...

AustinT:
> > I completely forgot about not having torch/dgl/etc mentioned in pyproject at all 😅
> 
> Yeah, I'm not sure what the best setup would be here...

Not sure either. ChatGPT didn't give any better suggestions than just telling users to install it themselves, so let's maybe just keep it as is?

PR #111, opened 2024-11-28
Fix link to installation instructions in README

kmaziarz:
As mentioned in #106, one of the docs links in README was missing the part that specifies the version, and thus it defaulted to a stale version from before #77. This PR fixes the link.


AustinT (approved):
Thanks Kris!

PR #113, opened 2024-11-28
Drop support for Python 3.7

kmaziarz:
Python 3.7 is well past its end-of-life; we mostly maintained support for it due to it being required to run GLN, which since #103 is no longer the case. This PR raises the minimum Python requirement to 3.8 and removes compatibility hacks that are not needed anymore. The recently released `v0.5.0` will thus be the last version to support Python 3.7, which users can fall back to if they really need to.


AustinT (approved):
LGTM, just remember to add a CHANGELOG entry.

PR #114, opened 2024-11-29
Fix search CLI crash under Python 3.11+

kmaziarz:
Some of the configuration options in the search CLI use mutable dataclasses as defaults; for some reason, this is only detected by Python 3.11 onwards, therefore our CI running under 3.7 - 3.9 did not flag it, but under Python 3.11 the search CLI crashes with an error. What's worse, one of our environment files (`environment.yml`) did not pin the version of Python, leading to 3.13 being selected, under which the error is present.

This PR makes the following changes:
- Use the `field` constructor instead of mutable defaults to restore proper behaviour under all Python versions.
- Run the CI under all Python versions ranging from 3.8 to 3.13 (the last stable version). In theory, it would likely be enough to test under 3.8 (oldest supported), 3.9 (used by single-step models) and 3.13 (newest supported). On the other hand, the core tests are fast and run in parallel, so testing under all versions shouldn't significantly increase time taken to merge PRs. Having more CI variants increases the chance that we hit some test flakiness and need to rerun a subset of the CIs, but if anything this will motivate us to fix the flaky tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


kmaziarz:
Actually, regarding the GitHub Actions limits: given that this repo is under a corporate `microsoft` account, and not a private one, I'm not sure what limit (if any?) applies here. Running CI in syntheseus does not seem to consume the free GitHub Actions minutes tied to my personal account.

AustinT (approved):
Looks good Kris, thanks!! If the CI usage becomes a problem in the future we can reduce it, but I wouldn't worry about it now.

PR #122, opened 2025-04-01
Query the underlying reaction model in a fixed order

kmaziarz:
The `ReactionModel` class wraps an underlying reaction model with extra functionality, notably caching. If caching is turned off, as is the case e.g. during single-step evaluation, the inputs to the higher-level class are directly passed into the model. However, due to the use of set comprehension, the order of inputs was not being preserved. This did not lead to incorrect behaviour, as later we assemble outputs in the right order again, but could be confusing during debugging, e.g. if the underlying model prints out logs for each reaction processed, those logs would appear in a scrambled order. The simple fix in this PR is to perform deduplication via `dict.fromkeys`, which was already being done in other parts of the codebase.


AustinT (approved):
Very good catch Kris, thanks for fixing this 🚀 

*EDIT*: remember to add a CHANGELOG entry though.

PR #124, opened 2025-05-13
Add a utility to load a `SmilesListInventory` from file

kmaziarz:
By far the most common way to provide a purchasable molecule inventory is `SmilesListInventory`, and the most common way of constructing that is from a file containing a list of SMILES, one per line. So far, the code to read the SMILES, strip endlines, and construct the inventory, lived in the search CLI. However, calling search from arbitrary Python I had to duplicate it. This PR avoids this (minor) duplication by making this code a `classmethod` of `SmilesListInventory`.

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -194,10 +194,8 @@ def run_from_config(config: SearchConfig) -> Path:
     )
 
     # Set up the inventory
-    with open(config.inventory_smiles_file, "rt") as f_inventory:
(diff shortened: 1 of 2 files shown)


AustinT (approved):
Content-wise, LGTM! (I didn't run it myself but can't see any clear errors)

I notice the tests don't pass though, and the error is `urllib.error.HTTPError: HTTP Error 403: SSL is required`. This seems to be related to downloading the models. Is this the first time you are seeing an error like this?

PR #126, opened 2025-06-03
Refactor search CLI to make it easier to access programmatically

kmaziarz:
Our search CLI has so far been structured in a way that makes it hard to call parts of it programmatically from outside; for example, the code to parse and reorganize search algorithm hyperparameters was only present inside `run_from_config`. This PR refactors things to extract this logic out, and also separates arguments configuring the search algorithm itself from other configuration options. Finally, it also changes the type of the `search_algorithm` argument from `str` to an appropriate `Enum`, which allows for failing slightly earlier when an unsupported algorithm has been selected. Changes are split into reasonable commits for easier reviewing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 ### Changed
 
 - Drop support for Python 3.7 ([#113](https://github.com/microsoft/syntheseus/pull/113)) ([@kmaziarz])
+- Refactor search CLI to make it easier to access programmatically ([#126](https://github.com/microsoft/syntheseus/pull/126)) ([@kmaziarz])
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


AustinT:
> Using a string algorithm name for calling search programmatically still works [...]
>
> Programmatically calling CLIs using the config classes is still quite clunky and not very intuitive, because even though type hints suggest you pass the config classes directly to `run_from_config`, this is just not true, you need to map them through `OmegaConf` first, which actually gives you a different class which has the same fields [...]
> 
> So, in summary, I think any usecase that worked before still works, except we do get an extra point of unhappiness for type checkers, but in this setting those were unhappy already due to the `model_class` field.

Ah I see, I forgot about the mapping through `OmegaConf` and thought that the mypy failure was caused by `str` types no longer being accepted. And I agree that if you are calling this programatically then you wouldn't be using the CLI anyway. I no longer have any objections towards using `Enum`, thanks for explaining everything clearly 😄

PR #127, opened 2025-06-24
Support LRU cache eviction for reaction models

kmaziarz:
Our base `ReactionModel` class implements simple caching, which avoids calling the model multiple times with the same input. However, there is no way to limit the growth of said cache, which can be problematic in longer-running deployments (e.g. server which runs indefinitely and accepts single-step or multi-step requests). This PR adds support for LRU-style cache eviction, which can be enabled by setting the `max_cache_size` parameter in the constructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Support LRU cache eviction for reaction models ([#127](https://github.com/microsoft/syntheseus/pull/127)) ([@kmaziarz])
 - Add a utility to load a `SmilesListInventory` from file ([#124](https://github.com/microsoft/syntheseus/pull/124)) ([@kmaziarz])
 
 ### Fixed
--- a/syntheseus/interface/models.py
+++ b/syntheseus/interface/models.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import warnings
 from abc import abstractmethod
+from collections import OrderedDict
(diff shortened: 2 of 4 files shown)


AustinT (approved):
LGTM! Only potential concern is around performance: I had considered `OrderedDict` initially for the caching but decided against because I read that it adds a lot of overhead (but I never actually tested it out to see how significant the overhead is). The overhead is _probably_ minimal compared to model inference, but nonetheless might be worth considering. Overall I'm happy for this to be merged without any changes though :)

kmaziarz:
> I had considered `OrderedDict` initially for the caching but decided against because I read that it adds a lot of overhead

I looked into this; it seems the concerns about `OrderedDict` being slow come from old Python. It used to be much slower than `dict` but [in 3.5 the implementation was changed](https://docs.python.org/3/whatsnew/3.5.html#whatsnew-ordereddict), bridging the gap almost entirely.

Empirically, on my machine, `10^6` insertions with an `int` key take 76ms for `dict` and 122ms for `OrderedDict`. If in each iteration I instead make a random `Molecule` for the key, then the time for both grows to ~8s, and there is no visible difference anymore, perhaps because constructing the `Molecule` and hashing it is taking up most of the time. This is with `canonicalize=False` and `make_rdkit_mol=False`; with these things turned on it's another 10x slower, and again without much difference between `dict` and `OrderedDict`. Either way, it seems the difference is small, gets even smaller with more complex keys, and will be completely trumped by other operations even a dummy single-step model would have to do, like constructing `Molecule`s 🙂

PR #130, opened 2025-07-09
Implement option to expand target even when purchasable

kmaziarz:
This PR implements an `expand_purchasable_target` option, which forces the search algorithm to expand the target molecule, even if it is already purchasable. This is similar to `exclude_target_from_stock` from [aizynthfinder](https://molecularai.github.io/aizynthfinder/configuration.html). The implementation is similar to the existing `expand_purchasable_mols`: purchasability metadata is still set according to the full inventory, but if the option is set then expansion happens regardless.

For the record, I also considered (but rejected) two alternative ways of implementing this feature:
- Taking `aizynthfinder`'s naming literally by temporarily removing the target from the inventory. Although an abstract inventory does not necessarily hold a collection of purchasable molecules inside, I considered adding `exclude(Molecule)` and `clear_exclusions()` methods to `BaseMolInventory` to handle these "temporary exclusions". However, this would make the inventory class mutable, which would go against the implicit assumption of immutability our search algorithms seem to make. It would also mean we'd have to be extra careful to remove the exclusions after search is done (e.g. when search is interrupted with an exception, which is then caught, we don't want the inventory to be left in a broken state).
- Overriding the purchasability metadata received from the inventory when the molecule is the target, to have it marked as non-purchasable, as if it was not in the inventory to begin with. This is inconsistent with the behaviour of `expand_purchasable_mols`, and would mean the metadata on the molecules may not agree with the state of the inventory.

### DIFF
--- a/syntheseus/cli/search.py
+++ b/syntheseus/cli/search.py
@@ -128,6 +128,7 @@ class SearchAlgorithmConfig:
     limit_graph_nodes: int = INT_INF
     prevent_repeat_mol_in_trees: bool = True
     stop_on_first_solution: bool = False
+    expand_purchasable_target: bool = False  # Whether to expand target even if it's purchasable
 
 
 def search_algorithm_config_to_kwargs(config: SearchAlgorithmConfig) -> Dict[str, Any]:
@@ -140,6 +141,7 @@ def search_algorithm_config_to_kwargs(config: SearchAlgorithmConfig) -> Dict[str
             "limit_graph_nodes",
             "prevent_repeat_mol_in_trees",
             "stop_on_first_solution",
+            "expand_purchasable_target",
         ]
     }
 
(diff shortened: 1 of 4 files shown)


AustinT:
> Not related to benchmarking; this arose when working towards on a real deployment, it was recommended by @mrwnmsr that we add this feature. Also, I'm not sure if I'm reading `aizynthfinder`'s code correctly, but it seems to me that their equivalent (`exclude_target_from_stock`) is actually on by default, would could further suggest this is what users may expect.

Ok interesting, I did not know that!

You make a good point about the downsides of my suggested implementation. I had envisioned it as a CLI-only flag if the main use case was benchmarking. Since you suggest this feature will be used outside of just benchmarking then let's stick with your implementation?

PR #146, opened 2026-01-27
Fix automatic checkpoint download

kmaziarz:
Recently, [our CI started failing](https://github.com/microsoft/syntheseus/actions/runs/21395087219/job/61612531978) due to the inability to download default model checkpoints. This also happens locally, and [has been reported online](https://www.reddit.com/r/bioinformatics/comments/1qg7666/figshare_downloads_blocked_by_aws_challenge/). Issue seems to be caused by the links (which are of the form `https://figshare.com/ndownloader/files/XXX`) no longer working for programmatic download due to an AWS security check. [Figshare API](https://info.figshare.com/user-guide/how-to-use-the-figshare-api/) doesn't directly discuss programmatic downloads, but mentions how to query the API to get entry metadata, and that contains a download link. Currently, this metadata points to links of the form `https://ndownloader.figshare.com/files/XXX`, which do work for downloading programmatically. Since it's not clear if the link format will stay stable or change in the future, in this PR I propose to get the link at runtime using Figshare API. One small complication is that the file IDs are separate from Figshare entry IDs, so the actual numbers recorded in the `*.yml` file all had to change.


AustinT (approved):
LGTM, thanks for the fix!

Curious if you think this has unintended security implications? Previously we were linking to a fixed file which IIUC could not be updated without a new link being issued. Now we retrieve metadata and get the link from that. Would it be possible for somebody to update the record to include compromised files, and that the metadata would link to the updated file, causing it to get automatically downloaded without us specifically choosing to update our model checkpoints?

If this is an issue it's not a blocker that should prevent it being merged IMO, just worth thinking if there is a workaround. I'm not aware of a workaround though.

mrwnmsr (approved):

PR #148, opened 2026-02-13
Filter reactions where one of the reactants is the product

jla-gardner:
This PR add an option to all reaction models to filter out reactions where the product is also one of the reactants. This is hidden behind a flag `remove_product_in_reactants`, which is `False` by default for backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


jla-gardner:
@microsoft-github-policy-service agree company="Microsoft"

AustinT (approved):
This makes sense to me. I vaguely recall thinking about this problem when initially implementing this library but forget why I didn't include a kwarg like this. Thanks for implementing!

kmaziarz (approved):
LGTM. Before merging, remember to update the CHANGELOG, and also fill in the PR description (see any other PR e.g. #144), which will then populate the commit message of the merge commit added to `main` 🙂

mrwnmsr (approved):

PR #149, opened 2026-02-17
Implement resumable single step evals

jla-gardner:
This PR (attempts to) makes the `syntheseus/cli/eval_single_step.py` script resumable, mitigating the impact of e.g. job crashes/pre-emption.

To do this, the script now saves model inputs+outputs+information in a [`JSON Lines`](https://jsonlines.org/) file during evaluation.  Re-invoking this script after e.g. pre-emption/restarts attempts to load this file if available, and skips calling the model for any already-queried inputs.

The new output format is easier to work with than the existing `stats.json` files, since you can load line-by-line (see attached script and analysis for what this looks like).

To preserve backwards compatibility, this feature is hidden behind the `config.resumable` flag, which is turned off by default.

Thank you to @kmaziarz for the suggested implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
-- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])  
+- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
 
 ## [0.7.2] - 2026-01-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Add a `resumable` flag when running single step predictions. By default this is set to False, preserving existing behaviour. If this is turned on, single step results for each input are saved to disk in a JSON-lines file, such that if a job is killed halfway through, running the same command will resume from the state the previous run reached. ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
+
 ## [0.7.2] - 2026-01-30
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
-- Implement resumability for single-step evaluations. ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
+- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])  
 
 ## [0.7.2] - 2026-01-30
 
@@ -170,3 +170,4 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 [@kmaziarz]: https://github.com/kmaziarz
 [@jagarridotorres]: https://github.com/jagarridotorres
 [@fiberleif]: https://github.com/fiberleif
+[@jla-gardner]: https://github.com/jla-gardner
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
-- Add a `resumable` flag when running single step predictions. By default this is set to False, preserving existing behaviour. If this is turned on, single step results for each input are saved to disk in a JSON-lines file, such that if a job is killed halfway through, running the same command will resume from the state the previous run reached. ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
+- Implement resumability for single-step evaluations. ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
 
 ## [0.7.2] - 2026-01-30
 
--- a/syntheseus/cli/eval_single_step.py
+++ b/syntheseus/cli/eval_single_step.py
@@ -216,7 +216,8 @@ class TimingRecord:
 class PredictionRecord:
     """Schema for a single row in the resumable predictions JSONL file."""
 
-    target: str
+    input: str
+    ground_truth: str
     num_predictions: int
     ground_truth_correct: List[bool]
     timing: TimingRecord
@@ -229,7 +230,8 @@ class PredictionRecord:
     def to_dict(self) -> Dict[str, Any]:
         """Serialize to a JSON-compatible dict, omitting unset optional fields."""
         d: Dict[str, Any] = {
-            "target": self.target,
+            "input": self.input,
+            "ground_truth": self.ground_truth,
             "num_predictions": self.num_predictions,
             "ground_truth_correct": self.ground_truth_correct,
             "timing": asdict(self.timing),
@@ -250,7 +252,8 @@ class PredictionRecord:
         """Deserialize from a JSON-loaded dict."""
         bt_timing = d.get("back_translation_timing")
         return cls(
-            target=d.get("target", ""),
+            input=d.get("input", ""),
+            ground_truth=d.get("ground_truth", ""),
             num_predictions=d["num_predictions"],
             ground_truth_correct=d["ground_truth_correct"],
             timing=TimingRecord(**d["timing"]),
@@ -509,8 +512,14 @@ def compute_metrics(
                         bt_preds = [
                             [rxn.reaction_smiles for rxn in seq] for seq in back_translation_results
                         ]
+                if isinstance(input, Molecule):
+                    input_smiles = input.smiles
+                else:
+                    assert isinstance(input, Bag)
+                    input_smiles = ".".join(mol.smiles for mol in input)
                 record = PredictionRecord(
-                    target=output.reaction_smiles,
+                    input=input_smiles,
+                    ground_truth=output.reaction_smiles,
                     num_predictions=len(reaction_list),
                     ground_truth_correct=ground_truth_matches,
                     timing=TimingRecord(
--- a/syntheseus/cli/eval_single_step.py
+++ b/syntheseus/cli/eval_single_step.py
@@ -36,8 +36,8 @@ from syntheseus.interface.models import (
     ReactionModel,
     ReactionType,
 )
-from syntheseus.interface.molecule import SMILES_SEPARATOR, Molecule
-from syntheseus.interface.reaction import REACTION_SEPARATOR, Reaction, SingleProductReaction
+from syntheseus.interface.molecule import Molecule, molecule_bag_to_smiles
+from syntheseus.interface.reaction import Reaction, SingleProductReaction
 from syntheseus.reaction_prediction.chem.utils import remove_stereo_information_from_reaction
 from syntheseus.reaction_prediction.data.dataset import (
     DataFold,
@@ -50,12 +50,8 @@ from syntheseus.reaction_prediction.inference.config import (
     BackwardModelConfig,
     ForwardModelConfig,
 )
-from syntheseus.reaction_prediction.utils.config import (
-    get_config as cli_get_config,
-)
-from syntheseus.reaction_prediction.utils.config import (
-    get_error_message_for_missing_value,
-)
+from syntheseus.reaction_prediction.utils.config import get_config as cli_get_config
+from syntheseus.reaction_prediction.utils.config import get_error_message_for_missing_value
 from syntheseus.reaction_prediction.utils.metrics import (
     ModelTimingResults,
     TopKMetricsAccumulator,
@@ -205,14 +201,6 @@ def get_results(
 
 
 @dataclass(frozen=True)
-class TimingRecord:
-    """Timing information for a single sample's model call."""
-
-    time_model_call: float
-    time_post_processing: float
-
-
-@dataclass(frozen=True)
 class PredictionRecord:
     """Schema for a single row in the resumable predictions JSONL file."""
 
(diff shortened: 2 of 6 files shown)


jla-gardner:
This script: [demo_resume.py](https://github.com/user-attachments/files/25369648/demo_resume.py) mocks the use of a simple model when performing single step evaluations, and shows the resumability of the process. 

The raw output is: 
```
{"input": "CN", "ground_truth": "C.N>>CN", "num_predictions": 3, "ground_truth_correct": [true, false, false], "timing": {"time_model_call": 2.1298726399739582e-05, "time_post_processing": 5.404154459635417e-06}, "predictions": ["C.N>>CN", "c1ccccc1>>CN", "NC=O>>CN"]}
{"input": "CN", "ground_truth": "NC=O>>CN", "num_predictions": 3, "ground_truth_correct": [false, false, true], "timing": {"time_model_call": 2.0265579223632812e-05, "time_post_processing": 2.9802322387695312e-06}, "predictions": ["C.N>>CN", "c1ccccc1>>CN", "NC=O>>CN"]}
{"input": "CC", "ground_truth": "C.C>>CC", "num_predictions": 3, "ground_truth_correct": [false, false, false], "timing": {"time_model_call": 2.0265579223632812e-05, "time_post_processing": 2.9802322387695312e-06}, "predictions": ["C.N>>CC", "c1ccccc1>>CC", "NC=O>>CC"]}
```

Parsing this into a more readable format: ` head -1 predictions.jsonl | python3 -m json.tool` gives:
```json
{
    "input": "CN",
    "ground_truth": "C.N>>CN",
    "num_predictions": 3,
    "ground_truth_correct": [
        true,
        false,
        false
    ],
    "timing": {
        "time_model_call": 2.1298726399739582e-05,
        "time_post_processing": 5.404154459635417e-06
    },
    "predictions": [
        "C.N>>CN",
        "c1ccccc1>>CN",
        "NC=O>>CN"
    ]
}
```

syntheseus/cli/eval_single_step.py
```diff
@@ -198,6 +204,91 @@ def get_results(
...
+            d["back_translation_predictions"] = self.back_translation_predictions
+        return d
+
+    @classmethod
+    def from_dict(cls, d: Dict[str, Any]) -> "PredictionRecord":
+        """Deserialize from a JSON-loaded dict."""
+        bt_timing = d.get("back_translation_timing")
+        return cls(
+            input=d.get("input", ""),
+            ground_truth=d.get("ground_truth", ""),
```
kmaziarz: Why would `input` or `ground_truth` be missing? Also, going in that direction, I'm wondering why we can't do something akin to
```python
return cls(
    **{
        k: TimingRecord(**v) if k.endswith("timing") else v
        for k, v in d.items()
    }
)
```
This would fail explicitly if extra stuff is provided in `d`, but maybe that's a good thing?

Same comment for `to_dict`; I suppose that also could be a for loop?

syntheseus/cli/eval_single_step.py
```diff
@@ -198,6 +204,91 @@ def get_results(
...
+            predictions=d.get("predictions"),
+            back_translation_predictions=d.get("back_translation_predictions"),
+        )
+
+
+def _reaction_from_smiles(rxn_smiles: str) -> Reaction:
+    """Reconstruct a Reaction from its reaction SMILES string."""
+    sep = REACTION_SEPARATOR * 2
+    reactants_str, products_str = rxn_smiles.split(sep)
+    reactants = Bag(Molecule(s) for s in reactants_str.split(SMILES_SEPARATOR) if s)
```
kmaziarz: We have `molecule_bag_from_smiles_strict` in `chem/utils.py` doing almost exactly this, modulo the rejection of empty SMILES components via `if s`. But I'm not sure why we'd want the latter?

More generally, maybe `_reaction_from_smiles` could be a `classmethod` of the `Reaction` class? I was somewhat surprised we don't have it there already.

syntheseus/cli/eval_single_step.py
```diff
@@ -198,6 +204,91 @@ def get_results(
     )
 
 
+@dataclass(frozen=True)
+class TimingRecord:
+    """Timing information for a single sample's model call."""
+
+    time_model_call: float
+    time_post_processing: float
```
kmaziarz: Is this not an exact copy of `ModelTimingResults` from `utils/metrics.py`? I guess the existing one is used for timing of a step (batch) and the new one you use for timing of a single sample (by dividing by batch size), but not sure if that warrants having both of them around.

syntheseus/cli/eval_single_step.py
```diff
@@ -356,10 +493,59 @@ def compute_metrics(
...
+                        else None
+                    ),
+                    back_translation_predictions=bt_preds,
+                )
+                jsonl_file.write(json.dumps(record.to_dict()) + "\n")
+                jsonl_file.flush()
+
         if include_predictions:
             all_predictions.extend(batch_predictions)
             all_back_translation_predictions.extend(batch_back_translation_predictions)
```
kmaziarz, line 499: So in resumable mode, are we flushing the predictions to a file as we go _and_ also keep all of them in memory? Or am I misunderstanding things?

syntheseus/cli/eval_single_step.py
```diff
@@ -356,10 +493,59 @@ def compute_metrics(
...
+                    )
+                    if include_predictions:
+                        bt_preds = [
+                            [rxn.reaction_smiles for rxn in seq] for seq in back_translation_results
+                        ]
+                if isinstance(input, Molecule):
+                    input_smiles = input.smiles
+                else:
+                    assert isinstance(input, Bag)
+                    input_smiles = ".".join(mol.smiles for mol in input)
```
kmaziarz: This is `molecule_bag_to_smiles` from `interface/molecule.py`.

kmaziarz (changes requested):
Thanks @jla-gardner, this has been annoying me since forever 🙂 I had an initial look and left a few comments, I can have another pass later. Before merging, I will also battle-test it on an actual problematic case (i.e. eval that spans multiple days yet gets pre-empted every 12h).

If you look at `CHANGELOG.md`, you'll find that we record all changes to `syntheseus` there (well, apart from those that don't affect any behaviour, e.g. changes purely in tests or fixing a typo in a comment). Feel free to add an entry there following the existing pattern (so, also append yourself to the contributors list)!

AustinT (commented):
Thanks for working on this @jla-gardner. The implementation roughly makes sense to me- I'll leave it to @kmaziarz for detailed review. I agree with his comments so far.

syntheseus/cli/eval_single_step.py
jla-gardner: Yes these methods were much to defensive I think - updated to be more compact as per your suggestion 😄

syntheseus/cli/eval_single_step.py
jla-gardner: Good point - I have added this as a class method. Interestingly, `ReactionSample` already had such a method, but (i) returned an `Optional[ReactionType]` and (ii) was not in fact used anywhere in the codebase. I've suggested removing this method to avoid the type clash.

syntheseus/cli/eval_single_step.py
jla-gardner: Agreed, and removed

syntheseus/cli/eval_single_step.py
jla-gardner: Yes this is correct: when `include_predictions=True` I'm proposing that we both flush each record to the JSONL file for crash resilience and keep the full predictions in memory: this function is expected to return the complete list of these predictions for summarisation into the final `stats.json` file, and I didn't want to change that behaviour here.

I think keeping things in memory should not be too large an overhead unless we are applying this to very large datasets?

syntheseus/cli/eval_single_step.py
```diff
@@ -262,6 +314,40 @@ def compute_metrics(
...
+                back_translation_timing_results.append(rec.back_translation_timing)
+            if include_predictions and rec.predictions is not None:
+                all_predictions.append([Reaction.from_reaction_smiles(s) for s in rec.predictions])
+            if include_predictions and rec.back_translation_predictions is not None:
+                all_back_translation_predictions.append(
+                    [
+                        [Reaction.from_reaction_smiles(s) for s in seq]
+                        for seq in rec.back_translation_predictions
+                    ]
+                )
```
kmaziarz: I think this can be sometimes costly because we're recreating a large amount of `Reaction`s, so a lot of `Molecule`s, thus lots of `rdkit` molecule objects, plus lots of SMILES re-canonicalization for no good reason (because we've saved them as canonical already). At least partially this is a general issue across `syntheseus`: we are making `rdkit`'s molecule objects happily all over the place (whenever a `Molecule` is made, unless you explicitly set `make_rdkit_mol=False`), where in fact usually they are not needed, and this blows things up memory-wise.

The "unneeded `rdkit` objects" side of things is a general pain-point, not specific to your changes; the re-canonicalization we could potentially address by having `from_reaction_smiles` take extra `**kwargs` which are passed to the `Molecules`s it creates, and then setting `canonicalize=False` when we load reactions from previously saved predictions.

syntheseus/cli/eval_single_step.py
kmaziarz: OK, makes sense, lets keep things as they are then. I do wonder if we want to rethink the larger design here and not have `EvalResults` store all results; this _can_ sometimes be an issue if your model returns a large volume of metadata per prediction (and all this is attached to the prediction objects). But that is orthogonal to the non-resumability issue you're addressing here so we can look at that separately.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Implement resumability for single-step evaluations. ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
```
kmaziarz: Nitpick for consistency with the rest of the file

```suggestion
- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
```

Also, I think you need to add yourself to the list at the bottom, otherwise `@jla-gardner` won't work (you can see [here](https://github.com/jla-gardner/syntheseus/blob/resumable-single-step-eval/CHANGELOG.md) how things are currently rendered).

syntheseus/reaction_prediction/data/reaction_sample.py
```diff
@@ -65,10 +65,3 @@ def from_reaction_smiles_strict(
             reagents=SMILES_SEPARATOR.join(sorted(reagents_smiles)),
             **kwargs,
         )
-
-    @classmethod
-    def from_reaction_smiles(cls: Type[ReactionType], *args, **kwargs) -> Optional[ReactionType]:
-        try:
-            return cls.from_reaction_smiles_strict(*args, **kwargs)
-        except Exception:
-            return None
```
kmaziarz: This is [used in RetroChimera](https://github.com/microsoft/retrochimera/blob/e97fac088a0c737d5b4970457a60214b42437dd2/retrochimera/data/processing.py#L26) 😆 The fact that we have it here yet don't actually use it anywhere in `syntheseus` is of course not ideal, and the whole setup of having a `Reaction` object _and_ a `ReactionSample` object is a result of merging two different abstraction hierarchies back in the day... We could look into refactoring things in the future but for now lets not remove this.

syntheseus/cli/eval_single_step.py
```diff
@@ -262,6 +314,40 @@ def compute_metrics(
...
+                ground_truth_match_modulo_stereo_metrics.add(rec.stereo_correct)
+            if back_translation_model is not None:
+                assert rec.back_translation_correct is not None
+                back_translation_metrics.add(rec.back_translation_correct)
+            num_predictions.append(rec.num_predictions)
+            model_timing_results.append(rec.timing)
+            if rec.back_translation_timing is not None:
+                back_translation_timing_results.append(rec.back_translation_timing)
+            if include_predictions and rec.predictions is not None:
+                all_predictions.append([Reaction.from_reaction_smiles(s) for s in rec.predictions])
```
kmaziarz: Now that I think of this round-tripping more carefully, I think we are losing `metadata` here (and `identifier`, but that one we never actually set to anything, it's a bit of a dangling abstraction that we thought we need but so far never actually used, whereas `metadata` contains all sorts of useful things).

The original format of saving reactions using `asdict_extended` does make for somewhat ugly outputs, but it records everything. Maybe we should keep using that in place of the (lossy) reaction SMILES?

kmaziarz: Now that I think of this round-tripping more carefully, I think we are losing `metadata` here (and `identifier`, but that one we never actually set to anything, it's a bit of a dangling abstraction that we thought we need but so far never actually used, whereas `metadata` contains all sorts of useful things).

The original format of saving reactions using `asdict_extended` does make for somewhat ugly outputs, but it records everything. Maybe we should keep using that in place of the (lossy) reaction SMILES?

syntheseus/reaction_prediction/data/reaction_sample.py
kmaziarz, line 74: This is [used in RetroChimera](https://github.com/microsoft/retrochimera/blob/e97fac088a0c737d5b4970457a60214b42437dd2/retrochimera/data/processing.py#L26) 😆 The fact that we have it here yet don't actually use it anywhere in `syntheseus` is of course not ideal, and the whole setup of having a `Reaction` object _and_ a `ReactionSample` object is a result of merging two different abstraction hierarchies back in the day. We could look into refactoring things in the future but for now lets not remove this.

kmaziarz (changes requested):

syntheseus/cli/eval_single_step.py
kmaziarz: (As noted in a separate comment, the fact that we have both `Reaction` and `ReactionSample` is confusing, but the method you found we do in fact use, just not in `syntheseus` itself)

syntheseus/cli/eval_single_step.py
jla-gardner: Very good catch! I updated to using this in the latest commit.

syntheseus/reaction_prediction/data/reaction_sample.py
jla-gardner: Oops! Reverted

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
```
kmaziarz: nit: trailing whitespace got added in the latest round of changes

```suggestion
- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
```

jla-gardner:
Updated to also persist reaction metadata and identifier. Example output: `head -1 predictions.jsonl | python3 -m json.tool`:

```json
{
    "input": "CN",
    "ground_truth": "C.N>>CN",
    "num_predictions": 3,
    "ground_truth_correct": [
        true,
        false,
        false
    ],
    "timing": {
        "time_model_call": 3.131230672200521e-05,
        "time_post_processing": 6.4373016357421875e-06
    },
    "predictions": [
        {
            "reactants": [
                {
                    "smiles": "C"
                },
                {
                    "smiles": "N"
                }
            ],
            "products": [
                {
                    "smiles": "CN"
                }
            ],
            "identifier": null,
            "metadata": {
                "source": "dummy_model",
                "reaction_id": 0,
                "ground_truth_match": true
            }
        },
        {
            "reactants": [
                {
                    "smiles": "c1ccccc1"
                }
            ],
            "products": [
                {
                    "smiles": "CN"
                }
            ],
            "identifier": null,
            "metadata": {
                "source": "dummy_model",
                "reaction_id": 1,
                "ground_truth_match": false
            }
        },
        {
            "reactants": [
                {
                    "smiles": "NC=O"
                }
            ],
            "products": [
                {
                    "smiles": "CN"
                }
            ],
            "identifier": null,
            "metadata": {
                "source": "dummy_model",
                "reaction_id": 3,
                "ground_truth_match": false
            }
        }
    ]
}
```

kmaziarz (approved):
This looks good now, thanks! There are a few extra tweaks needed to make this work reliably on Azure compute with mounted storage, but I can push these in separate smaller PR(s) once I fully work that out.

PR #151, opened 2026-04-24
Add abstractions for filtering backward model proposals

kmaziarz:
One important missing feature of syntheseus is support for filtering backward model proposals during search, for example by using a forward model or a feasibility model. This PR proposes some initial abstractions to do that, notably `ReactionFilterModel` to encapsulate all logic that filters proposed reaction steps. Given that filter and scoring models would also benefit from caching, I first extracted the caching-related logic from `ReactionModel` into a shared base class.

After we agree on these abstractions, the next step will be to add hooks into the search CLI to then enable basic forms of filtering during search, and log more statistics related to filtering itself (e.g. acceptance rates and fraction of time spent on filter model vs backward).

This PR is split into logical commits for easier reviewing.

### DIFF
--- a/syntheseus/reaction_prediction/filters/forward.py
+++ b/syntheseus/reaction_prediction/filters/forward.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import Sequence
 
 from syntheseus.interface.models import ForwardReactionModel, ReactionFilterModel
-from syntheseus.interface.reaction import Reaction
+from syntheseus.interface.reaction import SingleProductReaction
 
 
 class ForwardReactionFilterModel(ReactionFilterModel):
@@ -14,7 +14,7 @@ class ForwardReactionFilterModel(ReactionFilterModel):
         self.forward_model = forward_model
         self.top_k = top_k
 
-    def _get_acceptance(self, reactions: list[Reaction]) -> Sequence[bool]:
+    def _get_acceptance(self, reactions: Sequence[SingleProductReaction]) -> list[bool]:
         predictions = self.forward_model(
             [rxn.reactants for rxn in reactions], num_results=self.top_k
         )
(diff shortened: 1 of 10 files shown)


syntheseus/reaction_prediction/filters/forward.py
```diff
@@ -19,6 +19,6 @@ def _get_acceptance(self, reactions: Sequence[SingleProductReaction]) -> list[bo
             [rxn.reactants for rxn in reactions], num_results=self.top_k
         )
         return [
-            any(pred.product in rxn.products for pred in preds)
+            any(rxn.product in pred.products for pred in preds)
```
mrwnmsr, line 22: are there cases where bag membership could fail vs stricter equality?

kmaziarz: If you use a forward model that produces multiple products `r1.r2>>p1.p2`, then using bag membership guarantees both `r1.r2>>p1` and `r1.r2>>p2` are accepted. If we use equality, neither would be accepted.

If you use a forward model that produces always a single product, membership and equality testing would be equivalent.

mrwnmsr: might be worth documenting this, i can see cases for both, but also small footgun risk...

kmaziarz: What is the risk with the current approach i.e. bag membership?

mrwnmsr: somewhat corner case, but if we had a forward model that would generate a larger number of potential products and side products, this could slip through

mrwnmsr (approved):

AustinT:
@kmaziarz sorry for silence from me- last week was quite busy. I don't have time to look in detail now but the feature makes sense, thanks for adding!

PR #155, opened 2026-06-10
Improve resumable single-step eval

kmaziarz:
This PR continues after #149 by improving the robustness of single-step eval, especially in cases when the storage backend does not save things immediately (e.g. mounted remote container).

The most important change is that the current version of resumable evals rewrites the results file upon a restart, which is potentially a very heavy operation, and one that can take a minute or so to get flushed to storage. I saw cases where results from the next couple of batches would be lost because of a race condition between the file rewrite and subsequent appends. This PR proposes to instead truncate the file to remove the potential broken lines instead of fully rewriting it. Moreover, I found that reopening the results file after each batch instead of keeping it open throughout can lead to more frequent flushing to underlying storage. Finally, if running in resumable mode, `all_predictions` and `all_back_translation_predictions` were keeping the results in memory unnecessarily, as there is no downstream consumer of that data and it is already stored in the results file itself; this is also cleaned up here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
 
 ### Added
 
-- Implement resumability for single-step evaluations ([#149](https://github.com/microsoft/syntheseus/pull/149)) ([@jla-gardner])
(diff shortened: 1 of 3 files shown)


jla-gardner (approved):
LGTM!

PR #156, opened 2026-06-18
Integrate RetroChimera

kmaziarz:
This PR integrates RetroChimera and its submodels as possible single-step model choices. Among other things, this required updating the environment file with additional dependencies. As the models tests became more expensive to run, I've also included several changes designed to curb this somewhat; in particular, we used to test a full cartesian product of possible model and algorithm combinations, and this PR instead proposes to test each model once while cycling through search algorithms.

Changes are split into logical commits for easier reviewing.

### DIFF
--- a/environment_full.yml
+++ b/environment_full.yml
@@ -8,13 +8,20 @@ channels:
 dependencies:
   - faiss-gpu                                   # RetroKNN
   - numpy
+  - pandas                                      # RetroChimera
   - pip>=25.2
+  - pyg==2.5.2                                  # RetroChimera
   - python==3.9.7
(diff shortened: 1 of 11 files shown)


environment_full.yml
```diff
@@ -8,13 +8,20 @@ channels:
 dependencies:
   - faiss-gpu                                   # RetroKNN
   - numpy
+  - pandas                                      # RetroChimera
```
mrwnmsr, line 11: eventually, we need to replace pandas with polars 🐻‍❄️

mrwnmsr (approved):

PR #160, opened 2026-07-17
Disable sharded repodata in CI

kmaziarz:
Recently, our CI started failing from `main`, with errors in mamba env resolution. This turns out to be related to the handling of repodata information by the newest version of `mamba`. This PR disables the new behaviour, reverting to the old resolution method (whose only disadvantage is being slightly slower). We can remove this later once the new behaviour is stable.


fiberleif (approved):
lgtm

PR #163, opened 2026-07-31
Update new version release docs

kmaziarz:
In #158 we automated the workflow of releasing new versions to PyPI. However, `DEVELOPMENT.md` was not updated, and still contained the description of the more manual process; this is fixed in this PR. The text generally mirrors the [instructions in microsoft/retrochimera](https://github.com/microsoft/retrochimera/blob/main/DEVELOPMENT.md).


fiberleif (approved):
lgtm

